PR #102, opened 2013-05-23
pickle adapter

ethco:
I wrote a pickle adapter for recurly models. it's unfinished, but submitting it to you for feedback.


bhelx:
I don't think this needs to be part of the library as `pickle` is something not every project uses. It seems more like some kind of adapter you would put in your application. I'm going to close for now unless there are more people asking for pickle support.

PR #106, opened 2013-06-26
make Pager#each work more sensibly

ethco:
I've changed Pager#each, and thereby all other Enumerable methods of Pager, to behave the way that #find_each used to behave - that being the way <i>every other enumerable in ruby behaves</i>. 

when I do, say, `plan.add_ons.each`, I want to iterate over each add-on. not the first 50 add-ons, or any other first _n_ add-ons. I cannot imagine where I would want 50 add-ons but not the 51st. thus the former, page-specific behavior of #each is relegate to #each_current_page. 

I was bitten by records not being returned from #each - i.e., #each not behaving in a sensible and predictable manner, and hope to save others from this. 

this incidentally includes a fix for #find_each being broken and not yielding the same records each time it is called. as I do not know if this will prove too API-breaking, I will make a separate pull request for only that fix, but I am hoping the whole thing will be merged.

### DIFF
--- a/lib/recurly/resource/pager.rb
+++ b/lib/recurly/resource/pager.rb
@@ -2,7 +2,7 @@ module Recurly
   class Resource
     # Pages through an index resource, yielding records as it goes. It's rare
     # to instantiate one on its own: use {Resource.paginate},
-    # {Resource.find_each}, and <tt>Resource#{has_many_association}</tt>
+    # {Resource.each}, and <tt>Resource#{has_many_association}</tt>
     # instead.
     #
     # Because pagers handle +has_many+ associations, pagers can also build and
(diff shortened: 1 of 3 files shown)


lib/recurly/resource/pager.rb
```diff
@@ -64,24 +64,32 @@ def count
         @count ||= API.head(uri, @options)['X-Records'].to_i
       end
 
-      # @return [Array] Iterates through the current page of records.
+      # @return [Array,Enumerator] Iterates through the current page of records.
       # @yield [record]
-      def each
-        return enum_for :each unless block_given?
+      # Yields each record of the current page.
+      def each_current_page
```
sdhull: might consider naming this [`each_slice`](https://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-each_slice)

bhelx:
Closing due to age

PR #107, opened 2013-06-26
fix broken Pager#find_each

ethco:
this makes Pager#find_each return a consistent set of records each time it is called. it is included as part of #106 and can be ignored if that one is merged, but if that one does not make it, this one at least should.


bhelx:
Closing due to age

PR #116, opened 2013-09-24
add balance_in_cents_{invoiced,uninvoiced}

aheckmann:


bhelx:
As far as I can tell, this was never implemented server side. We can reopen if needed.

Issue #159, opened 2014-11-16
Threadsafe Recurly.api_key?


ivanovv:
We develop metrics app for Recurly, and extensively use threads (through Sidekiq). We have Braintree and Stripe integration in place and we’re working on getting Recurly live. But there is a problem - https://github.com/recurly/recurly-client-ruby/blob/master/lib/recurly/api/net_http_adapter.rb#L47 is not threadsafe, and two operations running in parallel will use the same api key (even if started for different clients).

Any chances to make Recurly thread safe?
Will you accept a PR?

thbar:
@ivanovv apart from this specific API key issue, since you appear to use Recurly in a multi-threaded setup, did you meet other thread-safety issues?

ivanovv:
@thbar we use [Sidekiq](https://github.com/mperham/sidekiq)  as a multithreaded background job solution, and no thread-safety issues for the last 6 months. But we mainly use this lib in readonly mode, so YMMV.

thbar:
@ivanovv thanks for the feedback, appreciated!

bhelx:
Closed with: https://github.com/recurly/recurly-client-ruby/pull/190

PR #161, opened 2014-12-01
Threadsafe Recurly.api_key

ivanovv:
Very straightforward solution, which brings a undesired behaviour - code, that relies on Recurly.api_key being set by some other thread,  will fail.
So, second commit to fix this issue, but for a main thread only - if you set Recurly.api_key in the main thread, it will be avail to all threads run afterwards. 
But if any of these threads sets its own key - it will use its own key.

### DIFF
--- a/lib/recurly.rb
+++ b/lib/recurly.rb
@@ -49,13 +49,17 @@ module Recurly
     # @return [String] An API key.
     # @raise [ConfigurationError] If not configured.
     def api_key
-      Thread.current[:recurly_api_key] or raise(
-          ConfigurationError, "Recurly.api_key not configured"
-      )
+      Thread.current[:recurly_api_key] or (defined? @api_key and @api_key) or raise(
+            ConfigurationError, "Recurly.api_key not configured"
+        )
     end
 
     def api_key=(api_key)
-      Thread.current[:recurly_api_key] = api_key
+      if Thread.main == Thread.current
+        @api_key = api_key
+      else
+        Thread.current[:recurly_api_key] = api_key
+      end
     end
 
     # @return [String, nil] A default currency.
--- a/lib/recurly.rb
+++ b/lib/recurly.rb
@@ -49,11 +49,14 @@ module Recurly
     # @return [String] An API key.
     # @raise [ConfigurationError] If not configured.
     def api_key
-      defined? @api_key and @api_key or raise(
-        ConfigurationError, "Recurly.api_key not configured"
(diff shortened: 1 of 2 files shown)


chrissrogers:
Related to issue recurly/recurly-client-ruby#159

chrissrogers:
I don't want to close this as it is an important feature for some. However, I am still hesitant to introduce `Thread.current` references to all users.

bhelx:
I agree. I think it's a design flaw. You should have to instantiate the api client given some credentials/subdomain as @chrissrogers mentioned. I'm also hesitant to solve it like this because we can't say for sure how it's going to work in everyone's environments and it would be a really difficult thing to debug.

Perhaps we can think of a way to support both using this current singleton API technique and also the instantiable way, then we can later deprecate the singleton way. I think I have a few ideas to do it. What do you think @chrissrogers ?

chrissrogers:
I support that path. It will take modifying all of the singleton calling patterns to on-instance calls. Basically every module will need to be instantiable and live on an instantiable Recurly. A singleton for backward compatibility will be easy to wrap around that.

ciddennis:
I created another pull request with my solution to this issue.  https://github.com/recurly/recurly-client-ruby/pull/190.

bhelx:
Well, after looking at doing an instantiable API object, it does seem like too big of a pattern change. I still want to get something in for these people.

I'm open to this solution as long as it doesn't affect people who are currently using accessing single accounts in a multithreaded environment. @ivanovv could you maybe add a test that proves this? Maybe something like this:

```ruby
    it "has the same config values unless a new one is set in the new thread" do
      Recurly.api_key = "123"
      Recurly.subdomain = "test"
      Recurly.default_currency = "USD"
      Thread.new do
        Recurly.api_key.must_equal "123"
        Recurly.subdomain.must_equal "test"
        Recurly.default_currency.must_equal "USD"
        Recurly.api_key = "456"
        Recurly.subdomain = "thread_test"
        Recurly.default_currency = "HKD"
        Recurly.api_key.must_equal "456"
        Recurly.subdomain.must_equal "thread_test"
        Recurly.default_currency.must_equal "HKD"
      end.join
      Recurly.api_key.must_equal "123"
      Recurly.subdomain.must_equal "thread_test"
      Recurly.default_currency.must_equal "USD"
    end
```

I don't think you are properly handling subdomains, default_currency, and the js config.

bhelx:
Fixed with https://github.com/recurly/recurly-client-ruby/pull/190

PR #173, opened 2015-01-09
Add Invoice#invoice_number_prefix and Invoice#invoice_number_with_prefix

cbarton:
cc/ @judith 

tests:

  - Turn on the Country Invoice Sequencing feature with VAT taxes
  - Create a VAT taxed invoice 
  - `Recurly::Invoice.find(<invoice_number_with_prefix>).invoice_number_with_prefix` should find the correct invoice and return its full invoice number.


judith:
:+1:

PR #175, opened 2015-01-21
Set `pending_subscription` currency to currency of parent subscription

whiteotter:
Currently `pending_subscription.currency` will set itself to the default currency for the site.  PR will allow subscriptions with non-default currencies to set the corresponding pending_subscription to the correct currency. 

Approvers: @bhelx


lib/recurly/subscription.rb
```diff
@@ -114,6 +114,11 @@ def subscription_add_ons= subscription_add_ons
     end
     alias add_ons= subscription_add_ons=
 
+    def pending_subscription
+      sub = self[:pending_subscription]
+      sub.tap {|e| e.currency = currency} if sub.is_a? Subscription
```
bhelx: Is there a reason you are using tap? Would this work in this situation?

```ruby
sub.currency = currency if sub.is_a? Subscription
```

whiteotter: done in that fashion, `subscription.pending_subscription` would only return the currency attribute for the pending_subscription

bhelx:
@whiteotter just a reminder, please add a changelog entry

PR #185, opened 2015-03-31
Define collection_method attribute in invoice

scottsherwood:
As defined in the docs at https://docs.recurly.com/api/invoices, adding collection_method attr to invoice resource.


cbarton:
This looks good, can you also add the `net_terms` attribute as well for invoices and then squash?

PR #190, opened 2015-05-07
Threadsafe Recurly.api_key?

ciddennis:
Resolves issue #159 Threadsafe Recurly.api_key?  by adding a new method that will allow developers to optionally override keys on a per thread context.

I know this is a lot like the current pull request for this issue. But this takes it a step further and allows any of the static config variable to be overridden.  Also it will only have any impact if the developer calls the Recurly.config() method.  Otherwise the normal code path stays in place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,21 @@ Recurly::API.net_http = {
 }
 ```
 
+## Multi-Threaded Configuration
+If you are using the client in a multi-threaded environment and require a different configuration per thread you can use the following within the threads context:
+
+``` ruby
+Recurly.config({
+  subdomain: ENV['RECURLY_SUBDOMAIN']
+  api_key: ENV['RECURLY_API_KEY'],
+  default_currency: "US",
+  private_key:  ENV['RECURLY_PUBLIC_API_KEY']
+})
+```
+
+The any configuration items you do not include in the above config call will be defaulted to the standard configuration items.   For example if you do not define default_currency then
+Recurly.default_currency will be used.
+
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Recurly::API.net_http = {
 ```
 
 ## Multi-Threaded Configuration
-If you are using the client in a multi-threaded environment and require a different configuration per thread you can use the following within the threads context:
+If you are using the client in a multi-threaded environment and require a different configuration per thread you can use the following within the thread's context:
 
 ``` ruby
 Recurly.config({
@@ -81,7 +81,7 @@ Recurly.config({
 })
 ```
 
-The any configuration items you do not include in the above config call will be defaulted to the standard configuration items.   For example if you do not define default_currency then
+Any configuration items you do not include in the above config call will be defaulted to the standard configuration items.   For example if you do not define default_currency then
 Recurly.default_currency will be used.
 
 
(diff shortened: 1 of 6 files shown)


ciddennis:
Sorry about the extra commits but find I also need to set different keys in the case of JS.

bhelx:
Thanks for the contribution. The code and the tests look pretty good here but I was actually thinking it would be better to keep the interface the same as it's done here: https://github.com/recurly/recurly-client-ruby/pull/161

Can you do it without the Recurly.config method? Or can you explain why you think this is better?

ciddennis:
I used the config as a way to stay out of the current code path as much as possible.  Basically it only comes in to play if the developer knows they are using threads and needs to have different configs based on a new thread.    If they never set the config then the existing variables are used to get the values.

In the end the most used path, Single Config across one or more threads, stays the same and only the new case, Multiple Config across multiple threads, alters the code path.

The other benefit to my way is all of the config values are covered not just the API Keys and if needed part of the config can be changed per thread while depending on the defaults for the rest of the config.

bhelx:
@ciddennis I like the explicit opt in aspect. I'm more concerned about the safety of the PR than I am having a single interface so I can get behind that approach.

A couple things I'd like to see before I can merge:

1. Add a note in the Configuration section of the README that explains how to use this and why you would want to use this. It can be pretty simple. Just needs to explain that you can scope different configuration options to the current thread if you are working with multiple credentials.
2. Add a bullet point to the Unreleased section of the changelog: https://github.com/recurly/recurly-client-ruby/blob/master/CHANGELOG.md

And one other thought I'd like to run by you. What do you think about possibly changing the method name from `config` to something that a little more descriptive to what it's actually doing? Maybe something like `thread_scoped_config`? I could go either way just wanted to know your opinion.

Thanks again for taking the time to work on this.

ciddennis:
Sure I can make the changes.  It might not be until next week.  Also the method name change will be fine.

Thanks

----------------------------------------------------------------------------------
Cid Dennis
cid.dennis@gmail.com
303-915-9381

bhelx:
No worries. Thanks.

bhelx:
@ciddennis have you had time to look at this again?

ciddennis:
Sorry I have not but will have time tomorrow.  I will get that completed so that this can be closed off.


Thanks

----------------------------------------------------------------------------------
Cid Dennis
cid.dennis@gmail.com
303-915-9381

ciddennis:
Ok I think I have the docs updated.

README.md
```diff
@@ -69,6 +69,21 @@ Recurly::API.net_http = {
...
+``` ruby
+Recurly.config({
+  subdomain: ENV['RECURLY_SUBDOMAIN']
+  api_key: ENV['RECURLY_API_KEY'],
+  default_currency: "US",
+  private_key:  ENV['RECURLY_PUBLIC_API_KEY']
+})
+```
+
+The any configuration items you do not include in the above config call will be defaulted to the standard configuration items.   For example if you do not define default_currency then
```
bhelx: `If there are` any configuration items...

README.md
```diff
@@ -69,6 +69,21 @@ Recurly::API.net_http = {
 }
 ```
 
+## Multi-Threaded Configuration
+If you are using the client in a multi-threaded environment and require a different configuration per thread you can use the following within the threads context:
```
bhelx: `thread's` context

bhelx:
Thanks @ciddennis

bhelx:
If you correct those 2 noted issues I can merge now

bhelx:
@ciddennis thanks so much for seeing it through!

PR #195, opened 2015-06-25
Use the klass that is instantiated associations when a Resource

cbarton:
When we build a `Resource` from its XML source, we are using the current classes associations, not the target class'. This will use the target class' associations when building the record.

cc/ @judith


judith:
:+1:

PR #199, opened 2015-07-13
Clarify required API Credentials

adarsh:
The documentation section for configuring the client library was missing a period which led to some confusion in setup.

This change adds markdown bullets to separate the two different required credentials.


bhelx:
Thanks @adarsh!

PR #207, opened 2015-08-21
Adding setup_fee_accounting_code

csmb:
Exposing a `setup_fee_accounting` code to the client library since it's now available.

Also updated spacing and fixed a typo in the Readme, and made the spec for creating a plan a bit more robust.


bhelx:
:+1:

Thanks for the extra effort

PR #218, opened 2015-10-21
Fixing spec

csmb:
Fixes a spec related to an impossible to set `setup_fee_accounting_code`. That filed does not allow spaces.


bhelx:
:+1:

Issue #222, opened 2015-11-06
Error updating account email: The provided XML was invalid. Unacceptable tag <redemptions>


vivanov26:
Gem version: 2.4.8

Steps to reproduce:

```
a = Recurly::Account.find 'xxxx-xxxx-xxxx'
a.email = 'changed' + a.email
a.save
```

Exception stack trace:

```
E, [2015-11-06T03:07:49.769867 #14575] ERROR -- : The provided XML was invalid. Unacceptable tag <redemptions> (Recurly::API::BadRequest)
/usr/local/rvm/gems/ruby-2.0.0-p481/gems/recurly-2.4.8/lib/recurly/api/net_http_adapter.rb:100:in `request'
/usr/local/rvm/gems/ruby-2.0.0-p481/gems/recurly-2.4.8/lib/recurly/api.rb:64:in `put'
/usr/local/rvm/gems/ruby-2.0.0-p481/gems/recurly-2.4.8/lib/recurly/resource.rb:792:in `save'
/var/uuportal/prod-portal.unblock-us.com/2015-11-03_20-20-58/app/services/recurly_billing_service.rb:311:in `set_account_email'
.....
```

alexz70:
![email update](https://cloud.githubusercontent.com/assets/11528043/10999575/d8df9f90-846a-11e5-8f0f-6270f8e80419.jpg)

vivanov26:
Just to clarify, it seems that ```<redemptions/>``` element gets included into the XML PUT request when updating email value. Here's an example from our production log:

```
<account><account_code>xxxxx</account_code><email>xxxxx@xxxxx.xx</email><redemptions/></account>
```

csmb:
Hi there @vivanov26. I can't seem to reproduce this locally.

When I try to reproduce, the XML sent over is:

```
<?xml version="1.0"?>
<account>
  <account_code>my_account</account_code>
  <email>changedmy_account@email.com</email>
</account>
```

Is there anything special about the account that you're trying to update, or does this happen on each account? Any extra info you've got would be great!

vivanov26:
Hi @csmb, thanks for looking into this. This may be related to our Recurly account. To answer your earlier question, yes it happens for every user account.

I think I figured out the additional step that needs to be done to reproduce this error: invoke
```
a.redemptions
```
before 
```
a.save
```

I will send the full trace to @rachelquick as I don't want to share our test account API key in this GitHub ticket.

csmb:
[...]

PR #223, opened 2015-11-11
Don't set changed attributes when reading a resource

csmb:
We don't want to set changed attributes when reading a resource, we only want to do that when writing to one. This change will return the changes if they're actual changes, and set the object's attribute to an empty array if they're not.

This addresses https://github.com/recurly/recurly-client-ruby/issues/222

cc: @bhelx


bhelx:
:+1:

PR #226, opened 2015-12-09
fix add_on quantity handling when adding an already existing add_on

Korrigan:
Hi,

I'm experiencing a weird behaviour when adding add-ons with quantity specified to subscription already having this add-on.

I can see in the Subscription::AddOns class that added add-on quantity is not handled properly if the add-on already exist since the code just adds 1 to the existing quantity.

I wrote a fix for this, I may write a test case too if you like.

What do you think?

Best regards,

### DIFF
--- a/lib/recurly/subscription/add_ons.rb
+++ b/lib/recurly/subscription/add_ons.rb
@@ -27,7 +27,7 @@ module Recurly
 
         exist = @add_ons.find { |a| a.add_on_code == add_on.add_on_code }
         if exist
-          exist.quantity ||= 1 and exist.quantity += 1
+          exist.quantity ||= 1 and exist.quantity += add_on.quantity || 1
 
           if add_on.unit_amount_in_cents
             exist.unit_amount_in_cents = add_on.unit_amount_in_cents


bhelx:
Hey @Korrigan

Would be happy to look at the code. Thanks for reporting! Could you also maybe show some code demonstrating the problem so I can verify it?

lib/recurly/subscription/add_ons.rb
```diff
@@ -27,7 +27,7 @@ def << add_on
 
         exist = @add_ons.find { |a| a.add_on_code == add_on.add_on_code }
         if exist
-          exist.quantity ||= 1 and exist.quantity += 1
+          exist.quantity ||= 1 and exist.quantity += add_on.quantity || 1
```
bhelx: This line already had poor readability. Could we perhaps break it up a little since we are in here? I know this was already in there, but can we also change `and` for `&&`?

bhelx: Do you understand what the original intention of the `and` was here? As far as I can tell this should just be on two lines.

Korrigan: Hi,

I think the oneliner works as expected but maybe I haven't tested all the cases. :)
I'll refine this to use the 2 lines versions which is correct for sure.

Korrigan:
Hi @bhelx,

Yep, I'll provide an example and a test case showing the issue tomorrow.

Best regards,

bhelx:
@Korrigan Thanks

bhelx:
@Korrigan You still working on this? Need any advice on the test case?

PR #230, opened 2016-01-29
Set id attribute for transactions from webhooks

lmnleaf:
Description: Adds an id attribute to transactions so it can be used for transaction objects built from the xml sent in a webhook

To recreate the bug: 

```ruby
webhook_notification = Recurly::Webhook.parse(request.body.read)
transaction_id = webhook_notification.transaction.id
```

Results in the following error: 
NoMethodError: undefined method `id' for #<Recurly::Transaction:0x007fde896c7888>.

Testing the changes/fix: 
* use the parse method to parse a webhook sent for a new transaction
* call transaction and id on the webhook notification to get the uuid of the transaction

```ruby
webhook_notification = Recurly::Webhook.parse(request.body.read)
transaction_id = webhook_notification.transaction.id
```

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -316,6 +316,12 @@ module Recurly
       # @example
       #   Recurly::Account.find "heisenberg"
       #   # => #<Recurly::Account account_code: "heisenberg", ...>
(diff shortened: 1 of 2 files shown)


csmb:
+1

PR #231, opened 2016-02-17
Add currency attribute to BillingInfo

bhelx:
This allows the merchant to send a custom currency on billing info creation and update for verification purposes. For instance, if your site default is USD and you are trying to verify for EUR, you may end up having two verifications. 

### Testing

```ruby
account = Recurly::Account.find('1')

account.billing_info = {
  first_name: 'Verena',
  last_name: 'Example',
  number: '4111-1111-1111-1111',
  verification_value:  '123',
  month: 11,
  year: 2019,
  currency: 'EUR'
}

account.billing_info.save!
```

Approvers: @csmb

PR #232, opened 2016-02-19
Bump 2.5.1

bhelx:
* Add `currency` attribute to `BillingInfo` object so client can pass currency on create/update [PR](https://github.com/recurly/recurly-client-ruby/pull/231)


csmb:
+1

PR #233, opened 2016-04-08
Remove Gemfile.lock

pmorton:
Hi - I was getting ready fork this Gem and make some updates. I noticed that bundler would no run because the Gemfile.lock is out of date for the version of ruby I am running. My general practice for public gem ```Gemfile.lock``` files is to follow http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/. Specifically:

```
When developing a gem, use the gemspec method in your Gemfile to avoid duplication. In general, a gem's Gemfile should contain the Rubygems source and a single gemspec line. Do not check your Gemfile.lock into version control, since it enforces precision that does not exist in the gem command, which is used to install gems in practice. Even if the precision could be enforced, you wouldn't want it, since it would prevent people from using your library with versions of its dependencies that are different from the ones you used to develop the gem.

When developing an app, check in your Gemfile.lock, since you will use the bundler tool across all machines, and the precision enforced by bundler is extremely desirable for applications.
```

Hence this pull request removes the Gemfile.lock. Please consider taking this pull request.


pmorton:
Closing in favor of adding more rubies since this change is really a requisite.

bhelx:
Closed due to: https://github.com/recurly/recurly-client-ruby/pull/234

PR #234, opened 2016-04-08
Add more rubies to the testing matrix

pmorton:
Adding jruby-9000, ruby-2.2 and ruby 2.3


pmorton:
@bhelx I am thinking of merging the Gemfile.lock change into this change. Doing so will allow the ruby dependencies to float and autofix this issue. Any thoughts?

bhelx:
Sure thing, I'll check this out this afternoon.

pmorton:
@bhelx Great. It does look like there are some test failures on jruby19 and ruby 2.3. 

As you can see in #233 jruby19 fails when the dependencies float, so clearly there is a dependency issue on jruby 1.9. Do you have any insight on this.

Ruby 2.3 has some failures. I will try to do some digging in.

pmorton:
@bhelx All builds are passing now on all ruby versions. Github is having some fun with my squash, the two other commits referenced prior to mine are and artifact of my rebase and do not functionally change anything. Those commits are already in master

spec/recurly/api/errors_spec.rb
```diff
@@ -33,15 +33,15 @@
       end
 
       describe "when response body is empty string" do
-        before { response.expect :body, '' }
+        before { 2.times { response.expect :body, '' } }
```
bhelx: @pmorton can you explain why this was needed?

lib/recurly/xml/nokogiri.rb
```diff
@@ -46,7 +46,7 @@ def text= text
       end
 
       def to_s
-        root.to_xml(:indent => 0).gsub(/$\n/, '')
+        root.to_xml(:indent => 0,:save_with => Nokogiri::XML::Node::SaveOptions::NO_EMPTY_TAGS).gsub(/$\n/, '')
```
bhelx: @pmorton can you explain why you needed to add this option?

.gitignore
```diff
@@ -11,4 +11,7 @@ doc/*
 bin/*
 !bin/recurly
 
+#Gems should not checkin lock files
+Gemfile.lock
```
bhelx: What do you think about referencing the ykatz article here in the comment so we can remember our reasoning later on?

http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

lib/recurly/resource.rb
```diff
@@ -401,7 +407,12 @@ def from_xml xml
             }
             next
           end
-
+          
+          # Nokogiri on Jruby-1.7.19 likes to throw NullPointer exceptions
+          # if you try to run certian operations like el.attribute(''). Since
+          # we dont care about text nodes, let's just skip them
+          next if el.node_type == 3
```
bhelx: This seems like a magic number. Is there possibly a named constant in nokogiri we can reference?

lib/recurly/xml/nokogiri.rb
pmorton: Silly Jruby-1.9. All other ruby flavors in the matrix generate an empty tag (a tag with no content) as follows ```<element></<element>```. Jruby 1.7.19 (against a floating nokogiri) generates ```</element>```. When checked against the expectation it does not match because jruby 1.7.19 is using self closing tags. This directive tells nokogiri to not generate self closing tags.

.gitignore
pmorton: 👍

spec/recurly/api/errors_spec.rb
pmorton: I am not a MiniTest expert, however as best as I can tell, old versions of minitest would simply use the same response for multiple calls to the same mock. It seem that in later versions you need to specify the response for each expected call in the expected order (http://stackoverflow.com/questions/26806274/multiple-invocations-of-same-method-on-a-mock-with-minitest). In this specific case the call issues at https://github.com/recurly/recurly-client-ruby/blob/master/lib/recurly/api/errors.rb#L49 will result in two calls to .body the first call could be services and the second would result in an error because the expectation did not know how to respond.

bhelx:
@pmorton thanks for being so responsive! I'm going to have to pick this back up this weekend or first thing Monday morning.

spec/spec_helper.rb
```diff
@@ -23,9 +24,26 @@ def stub_api_request method, uri, fixture = nil
...
 end
+
 include SpecHelper
 
+environment
+
+class Minitest::Spec
+  after do |tests|
+    WebMock.reset!
+    environment
```
bhelx: Why is this also called on line 38? Could you explain why this is needed?

bhelx: @pmorton if you can help me understand what this does I should be able to merge today :)

pmorton: Looks like line 38 can be killed and I will do so. Reading your question, the method call clearly is not expressive enough. It should be called ```reset_recurly_environment```. The purpose of the call is initialize the environment to a know state. The source of the need for this was two fold. In my local environment the tests were a bit flakey in that some times they would pass and others they would not. I added the after block to do two things:

1. Cleanup WebMocks - Each time a request was being stubbed, it would bleed over into other tests. I.e. Mocks created in one test were available in other tests. This behavior creates a lack of clarity and in this case some flaky tests. 

2. Environment - The recurly gem uses global state to store things like API tokens, domains, etc. There were some test that were resetting the global state in order to test the behavior of the global state. As a matter or practice, I thought that it was important to make sure that each test (even if it is not twiddling global state) reset the global state to a know good value after the completion of the test. Doing so eliminates an entire set of spec bugs that are slightly difficult to diagnose and ad in this case will cause tests to fail depending on the order that they are executed.

bhelx: :+1: Okay that sounds good

.gitignore
```diff
@@ -11,4 +11,8 @@ doc/*
 bin/*
 !bin/recurly
 
+# Gems should not checkin lock files
+# http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/
```
bhelx: :+1:

pmorton:
@bhelx I have addressed your feedback, squashed the commit - the build is passing. Since this is close to merge (and it makes development easier) I am going to use this as the base for #235

bhelx:
@pmorton Thanks for being so responsive!

Issue #235, opened 2016-04-08
Autoloading and Rails


pmorton:
We are getting some strange exceptions from the recurly railtie each time we deploy our application. The app is rails 4.2 with puma running with 2 clustered mode workers with 16 threads. When the application is deployed the first call into our app (any controller) will trigger the following exception: ```NoMethodError: undefined method `accept_language=' for Recurly::API:Class``` and a massive stack trace in what looks like the request filter phase. This generally does not make sense because the method is defined. After much research, it seems that this is _probably_ an issue with the use of autoload. 

https://bugs.ruby-lang.org/issues/921 confirms that for 1.9.3 and older, autoload is not thread safe while 2.x is. Further @headius seems to indicate that autoload is only thread safe is the same autoload mechanism is used across the board. https://github.com/headius/thread_safe/issues/15#issuecomment-28929979. So here is where I believe that issue to be: Rails uses it's own autoloading systems, while this gem uses another. There is some race condition that is causing the ```Recurly::API``` class to only be partially loaded when it is called. 

To address this issue we could take one of a couple of approaches

1. Dump autoloading and use require. This would make the gem slower to load, but this should be trivial since the gem is pretty small.

2. Hook into the rails autoloading system. Each file using autoloads would add the following to the class
```extend ActiveSupport::Autoload if defined?(ActiveSupport::Autoload)```. The net result is that the activesupport autoloader which is used by rails would also be used by this gem. Doing this would allow lazy loading for ruby files in development and eager loading in production.

I believe that 1 is the simplest with very little performance tradeoff, but if load performance is important to you option 2 would work as well.  Let me know what you think. I am happy to code this up.

pmorton:
Good Monday @bhelx Let me know your thoughts on the above and I would be happy to get this moving.

bhelx:
Thanks @pmorton

Will be taking a look today

bhelx:
@pmorton 

Okay, from what I can tell, our Railtie is pretty old and this library was written when using threaded servers was less common. So it's no wonder there are some issues.

At first glance I would definitely say I want to dump autoload (option 1). I find messing with ruby's requires to be a little to opaque and I never liked that this library adopted autoload. I also agree that the performance penalty will be negligible. I'd be happy to look over your PR or help you get it working.

pmorton:
Great. I will get this moving. Cheers!

bhelx:
Closed with https://github.com/recurly/recurly-client-ruby/issues/236

[...]

PR #236, opened 2016-04-12
Remove autoload and reorder require

pmorton:
This pull request addresses #235. Autoload has been removed and the ```require``` statements have been reorder to support the required dependency order.

### DIFF
--- a/lib/recurly/error.rb
+++ b/lib/recurly/error.rb
@@ -1,0 +1,13 @@
+module Recurly
+  # The exception class from which all Recurly exceptions inherit.
+  class Error < StandardError
+    def set_message message
+      @message = message
+    end
+
+    # @return [String]
+    def to_s
+      defined? @message and @message or super
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/recurly/error.rb
```diff
@@ -0,0 +1,13 @@
+module Recurly
```
pmorton: I had to move ```error``` out into it's own file since it was defined in the file that was currently loading ```recurly.rb``` and after a ```require``` that depended on the class being defined. The options were to move this out to it's own file or to define the class before the require statement. IMHO doing the former was cleaner and more understandable. 

Paul

bhelx:
@pmorton Thanks for jumping on this so quick. I'm going to run through this as soon as I get a chance.

bhelx:
I'm hoping to get time to check this out this week :)

pmorton:
@bhelx BTW. We have been running this in production for the last 8 days. It resolved the missing constant/threading issue.

bhelx:
@pmorton I had some time to pull this down this morning. Everything looks good. It will go out in the next release.

Thanks again :+1:

PR #240, opened 2016-05-02
Use gemspec instead of gemfile

bhelx:
Let's use the gemspec rather than the gemfile. I'll add some more explicit versions for some of these after seeing how they behave in the test matrix.


bhelx:
need to get this working for jruby

PR #244, opened 2016-05-10
Parse Fraud object if it exists

bhelx:
If it's available, the transaction may contain a `fraud` object with some fraud information. This will parse that object into a Hash. I've aliased it as `fraud_info` because I find that is easier to understand and in line with the schema.

## Testing

* Make sure it prints an `int` score and a `string` decision and nothing more.

```ruby
t = Recurly::Transaction.find("35cc481437fa97b566ccf04f7bad4ae0")
puts t.fraud.inspect
#=> {"score"=>99, "decision"=>"APPROVE"}
```

### DIFF
--- a/lib/recurly/transaction.rb
+++ b/lib/recurly/transaction.rb
@@ -47,8 +47,10 @@ module Recurly
       tax_exempt
       tax_code
       accounting_code
+      fraud
     )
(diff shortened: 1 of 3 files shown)


csmb:
+1

Issue #247, opened 2016-05-23
Update the address for an account?


cionescu:
I am having issues updating the address for a Recurly Account. 

```ruby
$ a = Recurly::Account.find("[account-code]")
===> GET https://caring.recurly.com/v2/accounts/[account-code]
...
<=== 200 OK (216.6ms)
...
$ a.address
=> #<Recurly::Address address1: "330 Blah", address2: "Flat 506", city: "San Francisco", country: nil, phone: "4159365945", state: "CA", zip: "94158">
$ a.address.address1 = "New Address"
$ a.address.changed?
=> true
$ a.changed?
=> false
```

Then, if I try to save the account it doesn't make a request to recurly. Completely unrelated, but if I try to save the address I get a 404

```ruby
$  a.address.save
===> POST https://caring.recurly.com/v2/address
<=== 404 Not Found (589.4ms)
<errors>
  <error>No route matches [POST] "/v2/address"</error>
</errors>
```

Can you please let me know what's wrong with my approach?

drewish:
Does `a.save` work any better?

cionescu:
`a.save` returns true. But I would expect it to make a PUT request to recurly to update the account.

drewish:
does it update the account though?

cionescu:
No, it doesn't

bhelx:
I think the problem is that `account.changed?` is probably still false. Let me check this and confirm that this is a bug.

bhelx:
This doesn't look correct

```ruby
puts "address changed: #{account.address.changed?}"      #=> true
puts "address changed: #{account.address_changed?}"     #=> false
puts "changed attributes: #{account.changed_attributes}"   #=> {}
```

bhelx:
This is a quick fix: https://github.com/recurly/recurly-client-ruby/pull/248

But it's probably not the best solution moving forward. We're going to need to discuss how to better address this problem of `changed_attributes` with nested objects.

cionescu:
Thank you! And thanks for responding so quickly 😄

bhelx:
@cionescu Sure thing. I'm going to do a deeper dive tomorrow.

bhelx:
Closed with https://github.com/recurly/recurly-client-ruby/pull/248

Will go into release tomorrow

PR #248, opened 2016-05-23
Fix account.address_changed? dirty check

bhelx:
Fixed Issue https://github.com/recurly/recurly-client-ruby/issues/247

But this is only a quick fix. I think we need an audit and a conversation around all the problems with `changed_attributes` and nested models.

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -437,7 +437,9 @@ module Recurly
           else
             val = XML.cast(el)
             if 'address' == el.name && val.kind_of?(Hash)
-              record[el.name] = Address.new val
+              address = Address.new val
+              address.instance_variable_set(:@changed_attributes, {})
+              record[el.name] = address
             else
               record[el.name] = val
             end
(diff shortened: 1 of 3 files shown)


lib/recurly/resource.rb
```diff
@@ -437,7 +437,9 @@ def from_xml xml
           else
             val = XML.cast(el)
             if 'address' == el.name && val.kind_of?(Hash)
-              record[el.name] = Address.new val
+              address = Address.new val
+              address.instance_variable_set(:@changed_attributes, {})
```
bhelx: When the address is unmarshalled, it's fields are dirty. Not sure why. This is too hacky.

spec/recurly/account_spec.rb
```diff
@@ -181,6 +181,23 @@
...
+      Account.find 'abcdef1234567890'
+    end
+
+    it "should be dirty if address was modified" do
+      account.address.address1 = "1600 Pennsylvania Ave."
+      account.changed?.must_equal true
+    end
+
+    it "should be clean if address was not modified" do
+      require'pry';binding.pry;
```
drewish: ✂️  this

bhelx: oops

csmb:
+1

PR #254, opened 2016-06-23
Add gift card purchase, preview, & get

shulmang:
Adds support for gift card purchasing

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -472,7 +472,7 @@ module Recurly
       # @return [Association, nil] Find association for the current class
       #                            with resource class name.
       def find_association(resource_class)
-        associations.find{ |a| a.resource_class == resource_class }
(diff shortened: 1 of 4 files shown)


lib/recurly/resource.rb
```diff
@@ -127,7 +127,7 @@ class Resource
     #     e.message # => "Can't find Account with account_code = tortuga"
     #   end
     class NotFound < API::NotFound
-      def initialize message
+      def initialize(message)
```
bhelx: This is great, but is there any way you could put these changes in their own commit?

shulmang: Good call, @bhelx .  Should be all set.

bhelx: Thanks a lot. It's a minor thing but helpful for keeping blames clean.

bhelx:
Closing in favor of https://github.com/recurly/recurly-client-ruby/pull/257

PR #268, opened 2016-10-17
Cleanup code around message interpolation

bhelx:
Just a code cleanup. No functional changes.

@lmnleaf Let me know if this still works the same.

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -157,11 +157,8 @@ module Recurly
 
       def record_to_message
         @record.errors.map do |k, v|
-          if k == "base"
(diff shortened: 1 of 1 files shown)


sdhull:
IMO, you should rework [`Recurly::Resource::Errors#full_messages`](https://github.com/recurly/recurly-client-ruby/blob/master/lib/recurly/resource/errors.rb#L12) to do this, and then here simply call `@record.errors.full_messages`

bhelx:
@sdhull I'm open to doing that in a separate PR

lmnleaf:
+1

PR #273, opened 2016-10-25
adds geo_code to billing_info, account address, and shipping address

lmnleaf:
adds `geo_code` to `billing_info`, `account` level address, and `shipping_address`


bhelx:
:+1:

PR #275, opened 2016-11-09
Correct rails deprecation warning

jdeseno:
The `prepend_before_filter` method is being removed, this gem causes these warnings on Rails 5:

| DEPRECATION WARNING: prepend_before_filter is deprecated and will be removed in Rails 5.1. Use prepend_before_action instead.

Fix falls back if the replacement `prepend_before_action` is unavailable.

Apologies for lack of unit test; was unable to think of a way without stubbing/mocking.
Have confirmed change works on Rails 5.0.0.1.

### DIFF
--- a/lib/rails/recurly.rb
+++ b/lib/rails/recurly.rb
@@ -5,7 +5,12 @@ module Recurly
     end
 
     initializer :recurly_set_accept_language do
-      ActionController::Base.prepend_before_filter do
+      prepend_method = :prepend_before_filter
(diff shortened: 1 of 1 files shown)


bhelx:
Thanks for submitting. Travis is failing because of a dependency problem with jruby and ruby 1.9. I'll fix that first.

bhelx:
@jdeseno Okay, dealt with those errors. No need to make you rebase, I checked this locally. Thanks for contributing!

PR #276, opened 2016-11-09
Fix failing dependency for ruby 1.9.3

bhelx:
It seems like our testing dependencies can't be fixed to work with ruby 1.9.3 (which is to be expected as it was EOL in early 2015). For now, I'm just ignoring 1.9.3 and jruby in the travis matrix. [but jruby-9000 still being tested]

They should continue to work but we will no longer automatically test them. In the meantime I'm trying to hunt down the failure in the webmock gem (which still claims to support 1.9.3).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,5 @@ matrix:
   allow_failures:
     - rvm: jruby-19mode
       env: XML=nokogiri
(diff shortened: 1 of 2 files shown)


lmnleaf:
+1

PR #277, opened 2016-11-10
Add new dunning event notification

bhelx:
Adds the new webhook event for new dunning notification:

https://dev.recurly.com/page/webhooks#dunning-event-notification-starting-11102016


lmnleaf:
+1

PR #279, opened 2016-11-15
Add back ruby 1.9.3 and jruby

bhelx:
After talking to webmock maintainer, was instructed to lock `addressable` at a version before it lost 1.9.3 compatibility. This adds ruby 1.9.3 and jruby back to the testing matrix.


lmnleaf:
+1

PR #286, opened 2016-11-29
Fix redemptions on bulk coupons

### LINKED ISSUE #284: redeem bulk coupon
I can't redeem a bulk coupon code
```ruby
coupon = Recurly::Coupon.find(bulk_coupon_code)
coupon.redeem(billing_account)
```
the error 
```
NoMethodError: undefined method `new' for []:Array
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/recurly-2.7.4/lib/recurly/coupon.rb:76:in `redeem'
	from (irb):4
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.6/lib/rails/commands/console.rb:110:in `start'
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.6/lib/rails/commands/console.rb:9:in `start'
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.6/lib/rails/commands/commands_tasks.rb:68:in `console'
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.6/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
	from /home/sigo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/railties-4.2.6/lib/rails/commands.rb:17:in `<top (required)>'
	from bin/rails:4:in `require'
	from bin/rails:4:in `<main>'
```

I think the problem is that bulk coupon don't have pager of redemptions.

If someone give me some hint of how to solve that. I can try to propose a pull request with the correction.

bhelx:
Resolves #284

Bulk coupons do not contain the `<redemptions />` element thus the `redemptions.new` line was failing. We use the `redeem` action and should only rely on that because it's always there. I had to change the `embedded!` method so we can at least call new on embedded resources. It's perhaps worth rethinking how we do that. Those methods should be protected not private.

Here is a script you can use to check that this works.

```ruby
require 'securerandom'

# use a bulk coupon code you created
coupon_code = 'cm2016-bx50h76kv1'
account_code = SecureRandom.uuid

# create a throwaway ccount
Recurly::Account.create(account_code: account_code)

coupon = Recurly::Coupon.find(coupon_code)

# should now throw an error
redemption = coupon.redeem(account_code)
puts redemption.inspect
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 <a name="unreleased"></a>
 ## Unreleased
 
+- Fix coupon redemption bug on bulk coupons #284 [PR](https://github.com/recurly/recurly-client-ruby/pull/286)
+
 <a name="v2.7.4"></a>
 ## v2.7.4 (2016-11-17)
 
--- a/lib/recurly/coupon.rb
+++ b/lib/recurly/coupon.rb
@@ -73,7 +73,7 @@ module Recurly
         :currency     => currency || Recurly.default_currency
       }.merge(extra_opts)
 
-      redemption = redemptions.new(redemption_options)
+      redemption = Redemption.new(redemption_options)
 
       Redemption.from_response follow_link(:redeem,
         :body => redemption.to_xml
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -579,8 +579,8 @@ module Recurly
       end
 
       def embedded!(root_index = false)
-        private :initialize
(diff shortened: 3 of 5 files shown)


lmnleaf:
+1

PR #288, opened 2016-12-01
adds id attribute to usage

### LINKED ISSUE #272: Subscription Add Ons Usage
When getting the usage on a subscription add-on it doesn't give you the ID for the usage in the response (although you can see it in the logs). Other than saving the ID to your db when it's created there's no way to get the ID of the usage from the API, which makes it impossible using this library to update or delete a usage.

lmnleaf:
Resolves #272

Description: Adds an id attribute to usage.


bhelx:
:+1:

spq24:
awesome! Thanks!

PR #290, opened 2016-12-08
Make the travis badge reflect master only

bhelx:
The travis badge was showing the latest build which doesn't reflect that master is actually stable.


lmnleaf:
+1

PR #293, opened 2017-01-05
Add all_line_items

bhelx:
If there are more than 500 line items, there will be a link to create a line item pager:

```ruby
invoice = Recurly::Invoice.find '1012'

# this if checks to see if the link is present in the xml
# if so, your line items were truncated
if invoice.all_line_items.any?
  # use find_each or all.each to iterate through every adjustment
  invoice.all_line_items.find_each do |adjustment|
    puts adjustment
  end
end
```


lmnleaf:
+1

PR #298, opened 2017-02-14
Adding product_code to Transaction and Adjustment

bhelx:
`product_code` should be writable on Adjustment and writable on Transaction. This is for Vertex integration.


lmnleaf:
+1

Issue #303, opened 2017-03-01
Please update Wiki


victorhazbun:
The Readme.md documentation is not enough, please add more documentation.

Thanks.

bhelx:
Hey @victorhazbun 

Are you interested in API level documentation or a more example driven approach? The readme is mostly for configuration and setup details.

You can find some API level docs here: http://www.rubydoc.info/github/recurly/recurly-client-ruby
and you can find some examples here: https://dev.recurly.com/

Would be happy to update the README with more info but I need more specifics.

victorhazbun:
For example: how to handle exceptions correctly? Whats the difference 
between Recurly::Subscription.create! and Recurly::Subscription.create?

Thank you.

Benjamin Eckel wrote:

victorhazbun:
By the way I just saw http://www.rubydoc.info/github/recurly/recurly-client-ruby/ but there is no reference to that documentation from the Ruby repository, it will be nice to have it there so users know how to use the API. :)

bhelx:
@victorhazbun Adding that link in #305. Thanks for pointing that out. I have a branch running with lots of improvements to the documentation. In the mean time feel free to file any questions you have as issues and we'll gladly help you out.

victorhazbun:
How can I send query parameters? https://dev.recurly.com/docs/list-accounts-subscriptions

   account = Recurly::Account.find('1')
   account.subscriptions.find_each do |subscription|
     puts "Subscription: #{subscription.inspect}"
   end

bhelx:
@victorhazbun sorry i didn't see your last comment until now. You can pass in params to anything that returns a `Pager`. So in your case would look something like this:

```ruby
account = Recurly::Account.find('1')
# subscriptions created after yesterday at midnight
# remember datetimes on the server are UTC
account.subscriptions.find_each(begin_time: DateTime.new(2017,4,19)) do |subscription|
  puts "Subscription: #{subscription.inspect}"
end
```

jpwynn:
Using v 2.7.6 of the gem, whcih supports API v2.4, the docs for v2.4 says find_each supports begin_time and end_time.

But this still selects invoices from today (2017) downwards, NOT from mid-2016 downwards:

```
Recurly::Invoice.find_each(begin_time: DateTime.new(2016,5,1), end_time: DateTime.new(2016,6,1)) do |invoice|

No doubt there is something missing, which is WHY updating the wiki with snippets that work would be helpful.

[...]

PR #305, opened 2017-03-09
Add yard docs link

bhelx:
This point was brought up in #303 

I have a PR that focuses on improving these docs as we have not previously done a very good job of using them and keeping them up to date.


lmnleaf:
+1

PR #317, opened 2017-03-23
Ban nokogiri on dead rubies

### LINKED ISSUE #316: Upgrade Nokogiri To 1.7.1
The current version of `nokogiri` this gem is using has [known vulnerabilities](https://github.com/sparklemotion/nokogiri/issues/1615). Can we update to 1.7.1?

bhelx:
Resolves #316 

The purpose of this PR is to officially deprecate ruby versions 1.9 and 2.0 and no longer allow
an insecure version of nokogiri to be used. Looks like we will also need to deprecate jruby-1.7 as it reports as ruby version 1.9.

- Bumping nokogiri to 1.7.1
- Deprecating ruby 1.9 and 2.0 and only allowing rexml use on those versions
- Conditionally install nokogiri development dependency (for ruby >= 2.1)
- Allow failures in Travis for appropriate versions with nokogiri
- Update README with supported version information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,5 +18,9 @@ env:
   - XML=nokogiri
 matrix:
   allow_failures:
-    - rvm: jruby-19mode
+    - rvm: 1.9.3
+      env: XML=nokogiri
+    - rvm: 2.0.0
+      env: XML=nokogiri
+    - rvm: jruby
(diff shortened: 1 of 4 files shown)


nadavshatz:
+1

PR #319, opened 2017-04-03
Cardless Free Trial v2.6 changes

bhelx:
These are some changes to the cardless free trial feature going into v2.6


benliscio:
@bhelx when will 2.6 be released?

bhelx:
@benliscio we don't have an official release date yet. You can use the `api_version_2_6` branch or one of these branches and it will work. We just cannot guarantee that a breaking change won't happen. Only when we update the docs and merge to master is the API frozen.

PR #324, opened 2017-04-18
Remove X-Records header parsing

bhelx:
We are removing the X-Records header in api version 2.6 on paginating GET requests
and they only come back on HEAD requests. Now calling `Resource#count` or `Pager#count`
will always explicitly call the server with a HEAD request. This should maybe be considered a breaking change:

To get a count, you'll need to explicitly call count which will call the server on each call.

```ruby
params = {
  begin_time: Date.today.prev_month,
  order: :asc,
  sort: :created_at
}

count = Recurly::Transaction.where(params).count
puts count
```


bhelx:
Okay, squashed

PR #334, opened 2017-07-11
Rework the pagination so it's simpler to explain and support

bhelx:
It was pointed out that the `Pager#find_each` method still only accepts an int option (per_page) when it should accept a Hash (for filtering and sorting). It was also pointed out that there isn't a consistent way or recommendation for how to do pagination. I decided I could take this opportunity to remove some of the API around pagination and only leave behind the recommended ways of pagination. 

This PR does a number of things:

1. Changes the signature of `Pager#find_each` to accept an options hash
rather than just the per_page param.
2. Removes aliased methods for pagination `all`, `where`, `scoped`.
3. Updates the yarddoc.

## Pagination Examples

```ruby
# given some filtering and sorting params (optional)
opts = {
  begin_time: DateTime.new(2016,1,1),
  end_time: DateTime.new(2017,1,1),
  state: :collected,
  sort: :created_at,
  per_page: 25
}

# To iterate through resources without worrying about paging, use `find_each`
# and the library will paginate for you under the hood
Recurly::Invoice.find_each(opts) do |invoice|
  puts invoice.invoice_number
end

# Or call find_each on any pager you get from a relationship
account.invoices.find_each(opts) do |invoice|
  puts invoice.invoice_number
end

# To manually iterate through pages, use `paginate` to create the Pager
# and `each` to iterate through each resource in the page
invoices = Recurly::Invoice.paginate(opts)
begin
  invoices.each do |invoice|
     puts invoice.invoice_number
  end
  puts "fetching next page..."
end while invoices.next
```

### DIFF
--- a/lib/recurly/account_balance.rb
+++ b/lib/recurly/account_balance.rb
@@ -16,6 +16,6 @@ module Recurly
     # This object does not represent a model on the server side
     # so we do not need to expose these methods.
     protected(*%w(save save!))
-    private_class_method(*%w(all find_each first paginate scoped where create! create))
+    private_class_method(*%w(find_each first paginate create! create))
   end
 end
--- a/lib/recurly/purchase.rb
+++ b/lib/recurly/purchase.rb
@@ -145,6 +145,6 @@ module Recurly
     # This object does not represent a model on the server side
     # so we do not need to expose these methods.
     protected(*%w(save save!))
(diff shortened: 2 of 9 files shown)


bhelx:
Putting a `WIP` back on for now. Would like to discuss removing a lot of this metaprogramming as discussed in #350

bhelx:
Closing until we can pick this back up.

PR #363, opened 2018-02-26
Invoice refunds do not return InvoiceCollection

### LINKED ISSUE #362: Invoice refunds do not return invoice collections
Hi, I'm having a little trouble upgrading to 2.14.0

`Recurly::Invoice#refund_amount` is expecting an `<invoice_collection>` as per the stubbed API response here https://github.com/recurly/recurly-client-ruby/blob/master/spec/fixtures/invoices/refund-201.xml

but when I hit that endpoint I get a top level `<invoice>` which blows up when the gem parses it.

```
curl -H 'Accept: application/xml' \
     -H 'X-Api-Version: 2.10' \
     -H 'Content-Type: application/xml; charset=utf-8' \
     -u '[api key]' \
     -d '<invoice><amount_in_cents>25</amount_in_cents><refund_method>transaction_first</refund_method></invoice>' \
    https://[subdomain]/v2/invoices/:invoice_number
```

bhelx:
Resolves #362

### DIFF
--- a/lib/recurly/invoice.rb
+++ b/lib/recurly/invoice.rb
@@ -164,28 +164,28 @@ module Recurly
 
     # Refunds specific line items on the invoice.
     #
-    # @return [InvoiceCollection, false] A new refund invoice, false if the invoice isn't
+    # @return [Invoice, false] Invoice if successful, false if the invoice isn't
     # refundable.
     # @raise [Error] If the refund fails.
     # @param line_items [Array, nil] An array of line items to refund.
     # @param refund_method ["credit_first", "transaction_first"] The method used to refund.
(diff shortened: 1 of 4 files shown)


Sinetheta:
`reload` won't work for methods that return different resources in their payload. They need to go back to `self.class.from_response(`

https://github.com/recurly/recurly-client-ruby/commit/49fbe3dd54b7c8f2abdfe5c1666910a4bc05fd0c

bhelx:
@Sinetheta I've returned the code to it's original state. If you have time, let me know if it works for you. I'll get a release out today.

Sinetheta:
Thanks, refunds are back 👍

PR #367, opened 2018-03-28
Fix nokogiri security warnings

bhelx:
We are also now blocking use of insecure versions of nokogiri.


lib/recurly/xml.rb
```diff
@@ -84,17 +84,35 @@ def to_s
...
+    version = Gem::Version.new(Nokogiri::VERSION)
+
+    if version.segments.length == 3
+      major, minor, patch = version.segments
+    else
+      major, minor, patch, _pre = version.segments
+    end
+
+    if minor < 6
+      raise insecure_noko_msg
```
bhelx: Trying to decide whether we want to raise or warn here.

PR #386, opened 2018-05-30
Make currencies method public

gevann:
Previously the currencies method on Money was protected. However, the
eql? method depends upon the currencies method being public, resulting
in two instances of Money with the same attributes not to be considered
equal to one another.

This commit exposes currencies publicly to fulfill the dependencies in
the eql? method.


bhelx:
Thank you for catching this @gevann

PR #391, opened 2018-06-15
Custom Fields

stepchud:
Adds `Recurly::CustomField` model with associations:

- **Account** `has_many :custom_fields`
- **Subscription** `has_many :custom_fields`

*Updating Account Custom Fields*
```ruby
# get custom fields on Account
> account.custom_fields
=> [#<Recurly::CustomField name: "dog_breed", value: "Pompoo">, #<Recurly::CustomField name: "dog_dad", value: "Clownsworth">]
# add a custom field to the account
> account.custom_fields << Recurly::CustomField.new(name: "dog_mom", value: "Cinderella")
=> [#<Recurly::CustomField name: "dog_breed", value: "Pompoo">, #<Recurly::CustomField name: "dog_dad", value: "Clownsworth">, #<Recurly::CustomField name: "dog_mom", value: "Cinderella">]
# delete an existing field
> account.custom_fields[1].value = nil
=> nil
# or
> account.custom_fields[1].value = ""

irb(main):021:0> account.save
=> true
```

*Updating Subscription Custom Fields*
```ruby
# get custom fields on subscription
> subscription.custom_fields
=> [#<Recurly::CustomField name: "salesperson", value: "jeff">, #<Recurly::CustomField name: "contact", value: "sales_email@example.com">]
# find an existing field by name, update the value
> subscription.custom_fields.find { |cf| cf.name == "salesperson" }.value = "jeff sales"
=> "jeff sales"
# delete an existing field
> subscription.custom_fields[1].value = nil
=> nil
# or
> subscription.custom_fields[1].value = ""

irb(main):021:0> subscription.save
=> true
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -202,7 +202,7 @@ This release will upgrade us to API version 2.8.
 
 ### Upgrade Notes
 
-There are two breaking changes in this API version you must consider. 
+There are two breaking changes in this API version you must consider.
 
 #### Country Codes
 
--- a/lib/recurly/account.rb
+++ b/lib/recurly/account.rb
@@ -42,6 +42,9 @@ module Recurly
     # @return [Pager<CreditPayment>, []]
     has_many :credit_payments, class_name: :CreditPayment, readonly: true
 
+    # @return [[CustomField], []]
+    has_many :custom_fields, class_name: :CustomField, readonly: false
+
     # Get's the first redemption given a coupon code
     # @deprecated Use #{redemptions} instead
     # @param coupon_code [String] The coupon code for the redemption
@@ -148,6 +151,9 @@ module Recurly
       attrs = super
       if address.respond_to?(:changed?) && address.changed?
         attrs['address'] = address
+      end
+      if custom_fields.any?(&:changed?)
+        attrs['custom_fields'] = custom_fields.select(&:changed?)
       end
       attrs
     end
--- a/lib/recurly/custom_field.rb
+++ b/lib/recurly/custom_field.rb
@@ -1,0 +1,15 @@
+module Recurly
+  # Recurly Documentation: https://dev.recurly.com/docs/custom-fields
+  class CustomField < Resource
+
+    define_attribute_methods %w(
+      name
+      value
+    )
+
+    # include the name with the serialized field, even though it can't change
+    def xml_keys
+      attributes.keys
+    end
+  end
+end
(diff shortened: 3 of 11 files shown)


lib/recurly/custom_field.rb
```diff
@@ -0,0 +1,17 @@
+module Recurly
+  # Delivery objects are only used with GiftCards.
+  #
+  # Recurly Documentation: https://dev.recurly.com/docs/custom-fields
```
stepchud: meant to update this, not sure where it will live

lib/recurly/account.rb
```diff
@@ -42,6 +42,9 @@ class Account < Resource
     # @return [Pager<CreditPayment>, []]
     has_many :credit_payments, class_name: :CreditPayment, readonly: true
 
+    # @return [Pager<CustomField>, []]
+    has_many :custom_fields
```
bhelx: Although this works, I prefer to have it explicitly spelled out:

```ruby
has_many :custom_fields, class_name: :CustomField, readonly: false
```

lib/recurly/custom_field.rb
bhelx: This should be fine for now. I'll update it if it's different before release.

stepchud:
WIP: this still needs to support custom fields on purchases.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 <a name="unreleased"></a>
 ## Unreleased
 
+<a name="v2.15.5"></a>
+## v2.15.5 (2018-06-06)
+
+- API version 2.13
```
bhelx: You don't need to fill these out, I typically do before the release. It will likely be a minor bump anyway due to subscription terms feature.

lib/recurly/account.rb
```diff
@@ -42,6 +42,9 @@ class Account < Resource
     # @return [Pager<CreditPayment>, []]
     has_many :credit_payments, class_name: :CreditPayment, readonly: true
 
+    # @return [Pager<CustomField>, []]
```
bhelx: Sorry, one more small detail. Let's change this since it will never be wrapped in a pager. Should be `[[CustomField], []]`

lib/recurly/subscription.rb
```diff
@@ -20,6 +20,9 @@ class Subscription < Resource
     # @return [Pager<Redemption>, []]
     has_many :redemptions
 
+    # @return [Pager<CustomField>, []]
```
bhelx: We need to change this yardoc too

PR #397, opened 2018-07-24
[V3] Test more versions of ruby

drewish:
And fix an issue with 2.3 where a value like `2` is represented as a `Fixnum` rather than an `Integer` like in later versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: ruby
 rvm:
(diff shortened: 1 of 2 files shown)


aaron-suarez:
So Ruby 2.4+ seems to be fine. Do we need to support 2.3?

drewish:
Yeah it's still a supported version, so I don't see a reason not to figure why this is failing.

PR #399, opened 2018-07-30
Make changes necessary for updating an invoice from client

aaron-suarez:
Allow editing of an invoice from the client.

### DIFF
--- a/lib/recurly/invoice.rb
+++ b/lib/recurly/invoice.rb
@@ -86,6 +86,7 @@ module Recurly
       line_items
       transactions
       terms_and_conditions
+      vat_reverse_charge_notes # Only shows if reverse charge invoice
       customer_notes
       address
       net_terms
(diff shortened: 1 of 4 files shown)


lib/recurly/invoice.rb
```diff
@@ -201,6 +202,22 @@ def xml_keys
...
+    # Raises an error if attempting to create an invoice using this method.
+    #
+    # @return [true, false]
+    # @raise [RuntimeError] Raises error if you attempt to create an invoice.
+    # @example
+    #   invoice = Recurly::Invoice.find('1000')
+    #   invoice.po_number = '1234'
+    #   invoice.save # => true
+    def save
+      if !persisted?
```
bhelx: This is a good use case for `unless`. `unless` is like `if` but it negates the conditional:

```ruby
unless persisted?
  # do something
end
```

lib/recurly/invoice.rb
```diff
@@ -201,6 +202,22 @@ def xml_keys
...
+        raise "New invoices cannot be created using Invoice#save"
```
bhelx: We might want to choose an error class here, something like:

```ruby
raise BadRequest, "New invoices cannot be created using Invoice#save"
```

lib/recurly/invoice.rb
aaron-suarez: Good catch

lib/recurly/invoice.rb
```diff
@@ -201,6 +202,22 @@ def xml_keys
...
+    #
+    # @return [true, false]
+    # @raise [RuntimeError] Raises error if you attempt to create an invoice.
+    # @example
+    #   invoice = Recurly::Invoice.find('1000')
+    #   invoice.po_number = '1234'
+    #   invoice.save # => true
+    def save
+      unless persisted?
+        raise "Invoices can only be updated with Invoice#save. New invoices cannot be created using this method."
```
aaron-suarez: Changed error message to clarify that invoices can be updated but not created with this method. Could not use any existing error classes, and it didn't make sense to create a new one at this time.

bhelx: 👍

PR #405, opened 2018-08-10
Support account notes

### LINKED ISSUE #327: The Account class does not seem to support notes
https://dev.recurly.com/docs/list-account-notes

aaron-suarez:
Fixes #327. Client could not access account notes using `account.notes`. Requires an API change (notes href tag must be included in Account xml).


bhelx:
Blocked until we get the V2 `/notes` link change

PR #407, opened 2018-08-13
Update to API 2018-06-06

bhelx:
* Fix Client#delete argument error
* Upgrade to latest 2018-06-06 partner doc


lib/recurly/requests/subscription_change_create.rb
```diff
@@ -3,8 +3,8 @@ module Requests
     class SubscriptionChangeCreate < Request
 
       # @!attribute add_ons
-      #   @return [Array[SubscriptionAddOnCreate]] If you set this value you include all the add-ons and their quantities and amounts. The values you include will replace the previous values entirely.
-      define_attribute :add_ons, Array, {:item_type => :SubscriptionAddOnCreate}
+      #   @return [Array[String]] If you set this value you include all the add-ons and their quantities and amounts. The values you include will replace the previous values entirely.
+      define_attribute :add_ons, Array, {:item_type => String}
```
drewish: Wondering why this is treating the AddOns as an array of strings.

bhelx: It's possible that I hand-corrected this problem before. I'm going to take a look at it when I get some time today.

PR #408, opened 2018-08-13
Use Net-http-persistent for persistent connection

bhelx:
Should discuss some configuration and dependency management here.

Tested against this script:

```ruby
require 'recurly'
require 'benchmark'

API_KEY = 'KEY'
SITE_ID = 'subdomain-SUBDOMAIN'

@client = Recurly::Client.new(api_key: API_KEY, site_id: SITE_ID)

# Get first 30 ids
ids = @client.list_accounts(limit: 30).each_page.first.map(&:id)

measurement = Benchmark.measure do
  # Fetch 30 accounts
  # Obviously we could use `ids=` in list_accounts but this
  # is unoptimized for benchmarking purposes
  ids.each do |id|
    acct = @client.get_account(account_id: id)
  end
end
puts measurement
```

Without persistent:
`0.520000   0.040000   0.560000 ( 11.453849)`
With persistent:
`  0.390000   0.030000   0.420000 (  4.655093)`

With persistent, the wall-clock time is over halved.

### DIFF
--- a/lib/recurly/client/adapter.rb
+++ b/lib/recurly/client/adapter.rb
@@ -1,0 +1,46 @@
+require 'openssl'
+
+module Recurly
+  class Client
+    module NetHttpPersistentAdapter
+      protected
+      def configure_adapter(faraday)
+        faraday.adapter :net_http_persistent do |http| # yields Net::HTTP
+          # Let's not use the bundled cert in production yet
+          # but we will use these certs for any other staging or dev environment
+          unless BASE_URL.end_with?('.recurly.com')
+            http.ca_file = File.join(File.dirname(__FILE__), '../data/ca-certificates.crt')
+          end
+          http.open_timeout = 50
+          http.read_timeout = 60
+        end
+      end
+    end
+
+    module NetHttpAdapter
+      protected
+      def configure_adapter(http)
+        faraday.adapter :net_http do |http| # yields Net::HTTP
+          # Let's not use the bundled cert in production yet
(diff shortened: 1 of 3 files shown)


drewish:
Yeah https://github.com/drbrain/net-http-persistent/issues/80 makes me wonder if we want to pull in that dependency.

bhelx:
@drewish yeah it's odd. Not sure why the project doesn't have release notes. But perhaps we can use `2.9`? Also, it might make sense to support this as an optional adapter.

bhelx:
Notes: https://github.com/drbrain/net-http-persistent/blob/master/History.txt

bhelx:
@cbarton right, thinking that might be the best option. Maybe also let them provide the version of `net-http-persistent` they use as well.

cbarton:
Or just a symbol and we provide the adapter integration

lib/recurly/client/adapter.rb
```diff
@@ -0,0 +1,46 @@
...
+          end
+          http.open_timeout = 50
+          http.read_timeout = 60
+        end
+      end
+    end
+
+    module NetHttpAdapter
+      protected
+      def configure_adapter(http)
```
drewish: realizing this probably should have been `faraday`

bhelx: https://github.com/recurly/recurly-client-ruby/commit/bb46c8869defe88df593d4e890d5a9de73713550

PR #414, opened 2018-09-14
Add support for Low Balance Gift Card Notification

aaron-suarez:
Adds support for the new webhook


lib/recurly/webhook/low_balance_gift_card_notification.rb
```diff
@@ -0,0 +1,6 @@
+module Recurly
+  module Webhook
+    class LowBalanceGiftCardNotification < SubscriptionNotification
+    end
+  end
+end
```
bhelx: Nit: you want your EOF to be on the last line. Files should end with newline + EOF.

aaron-suarez: Good catch

PR #422, opened 2018-10-02
Add missing webhook properties to models

### LINKED ISSUE #336: Webhook properties are missing
It appears that certain webhook properties are missing from the models.  For example, if I try to access a parsed webhook object and attempt to consume the account.phone property, I get an error that the phone property does not exist on the account object.  Yet, if I inspect the body of the webhook, I find that it includes the phone property.  I'm using API version 2.7 and gem version 2.10.1 which seems to indicate compatibility from the release notes.  There are other missing webhook properties as well, but I think it's better to review everything.

aaron-suarez:
Fixes #336 

This PR adds all of the missing properties to the relevant models based on the documentation at https://dev.recurly.com/page/webhooks. It also adds a few tests to the webhook spec that I was using while testing out my changes.

### DIFF
--- a/spec/fixtures/webhooks/new-credit-payment-notification.xml
+++ b/spec/fixtures/webhooks/new-credit-payment-notification.xml
@@ -14,7 +14,7 @@
     <action type="symbol">payment</action>
     <currency>USD</currency>
     <amount_in_cents type="integer">3579</amount_in_cents>
-    <original_nvoice_number type="integer">2389</original_nvoice_number>
+    <original_invoice_number type="integer">2389</original_invoice_number>
     <applied_to_invoice_number type="integer">2390</applied_to_invoice_number>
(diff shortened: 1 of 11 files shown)


spec/fixtures/webhooks/new-credit-payment-notification.xml
```diff
@@ -14,7 +14,7 @@
     <action type="symbol">payment</action>
     <currency>USD</currency>
     <amount_in_cents type="integer">3579</amount_in_cents>
-    <original_nvoice_number type="integer">2389</original_nvoice_number>
+    <original_invoice_number type="integer">2389</original_invoice_number>
```
bhelx: nice catch

aaron-suarez: Thanks 👍

PR #424, opened 2018-10-04
Handle malformed XML errors gracefully.

scottyp-env:
Sometimes errors come back from Recurly as HTML, which makes REXML choke while trying to log the error, so the real error is lost to us. If we handle the error a bit more gracefully then we can log the original backtrace and error.

I did think about doing this in the `REXMLAdapter` module, but I thought that we don't always want to quietly swallow malformed XML exceptions - e.g. if the response is OK and we should have valid subscription XML but don't, then we should blow up.

Because this code is specific to REXML, I've wrapped it up so it won't run when running the test suite for nokogiri. nokogiri seems to accept HTML just fine and doesn't throw a parse error.

### DIFF
--- a/lib/recurly/api/errors.rb
+++ b/lib/recurly/api/errors.rb
@@ -47,7 +47,7 @@ module Recurly
       def xml
         @xml ||= begin
           XML.new(response.body) if response and response.body and not response.body.empty?
-        rescue REXML::ParseException
+        rescue Recurly::XML::ParseError
           nil
         end
       end
--- a/lib/recurly/api/errors.rb
+++ b/lib/recurly/api/errors.rb
@@ -47,6 +47,8 @@ module Recurly
       def xml
         @xml ||= begin
           XML.new(response.body) if response and response.body and not response.body.empty?
+        rescue REXML::ParseException
+          nil
         end
       end
     end
--- a/lib/recurly/xml.rb
+++ b/lib/recurly/xml.rb
@@ -1,5 +1,7 @@
 module Recurly
   class XML
+    ParseError = Class.new(StandardError)
+
     class << self
       def cast(el)
         # return nil if the `nil` attribute is present
--- a/lib/recurly/xml/nokogiri.rb
+++ b/lib/recurly/xml/nokogiri.rb
@@ -2,7 +2,9 @@ module Recurly
   class XML
     module NokogiriAdapter
       def initialize xml
-        @root = Nokogiri(xml).root
+        @root = Nokogiri::XML(xml) { |config| config.strict }.root
+      rescue Nokogiri::XML::SyntaxError
+        raise ParseError
       end
 
       def add_element name, value = nil
--- a/lib/recurly/xml/rexml.rb
+++ b/lib/recurly/xml/rexml.rb
@@ -5,6 +5,8 @@ module Recurly
     module REXMLAdapter
       def initialize xml
         @root = ::REXML::Document.new(xml).root
+      rescue REXML::ParseException
+        raise ParseError
       end
 
       def add_element name, value = nil
--- a/spec/recurly/api/errors_spec.rb
+++ b/spec/recurly/api/errors_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe Recurly::API::ResponseError do
   describe "#xml" do
-    let(:xml) { '<?xml version="1.0"?>' }
+    let(:xml) { '<?xml version="1.0"?><root/>' }
 
     describe "when response not assigned" do
       let(:error) { Recurly::API::ResponseError.new nil, nil }
@@ -48,15 +48,13 @@ describe Recurly::API::ResponseError do
         end
       end
 
-      if ENV.fetch('XML', 'rexml') == 'rexml'
-        describe "when using REXML and response body is not valid XML" do
-          let(:html) { "<html><body><hr></body></html>" }
+      describe "when using response body is not valid XML" do
+        let(:html) { "<html><body><hr></body></html>" }
 
-          before { 3.times { response.expect :body, html } }
+        before { 3.times { response.expect :body, html } }
 
-          it "must return nil" do
-            error.send(:xml).must_equal nil
-          end
+        it "must return nil" do
+          error.send(:xml).must_equal nil
         end
       end
     end
--- a/spec/recurly/api/errors_spec.rb
+++ b/spec/recurly/api/errors_spec.rb
@@ -47,6 +47,18 @@ describe Recurly::API::ResponseError do
           error.send(:xml).must_be_instance_of Recurly::XML
         end
       end
+
+      if ENV.fetch('XML', 'rexml') == 'rexml'
+        describe "when using REXML and response body is not valid XML" do
+          let(:html) { "<html><body><hr></body></html>" }
+
+          before { 3.times { response.expect :body, html } }
+
+          it "must return nil" do
+            error.send(:xml).must_equal nil
+          end
+        end
+      end
     end
   end
 end
--- a/spec/recurly/xml_spec.rb
+++ b/spec/recurly/xml_spec.rb
@@ -1,6 +1,14 @@
 require 'spec_helper'
 
 describe Recurly::XML do
+  describe ".initialize" do
+    describe "when malformed XML is passed in" do
+      it "raises a Recurly::XML::ParseError" do
+        proc { Recurly::XML.new('<unclosed-root>') }.must_raise Recurly::XML::ParseError
+      end
+    end
+  end
+
   describe ".filter" do
     it "must filter sensitive data only on number and verification_value" do
       [


aaron-suarez:
@scottyp-env I would like to try this out. Can you please provide an example of an error you've seen that comes back from Recurly as HTML?

scottyp-env:
Hi @aaron-suarez , thanks for getting back to me. The problem we've got is that the error only occurs intermittently, and when it has been occurring it's been blowing up our error reporting so we can't see what the actual bad message is.

We've monkey-patched the recurly client gem to try and log any bad messages we receive, but wouldn't you know it, since we've done that all the responses have been fine. As soon as we get a complete message I'll post it here.

In the meantime I can share with you the error that we've seen:

```
REXML::ParseException: Missing end tag for 'meta' (got "head")
Line: 22
Position: 1077
Last 80 unconsumed characters:

1  File "/usr/local/lib/ruby/2.4.0/rexml/parsers/baseparser.rb" line 341 in pull_event
2  File "/usr/local/lib/ruby/2.4.0/rexml/parsers/baseparser.rb" line 185 in pull
3  File "/usr/local/lib/ruby/2.4.0/rexml/parsers/treeparser.rb" line 23 in parse
4  File "/usr/local/lib/ruby/2.4.0/rexml/document.rb" line 288 in build
5  File "/usr/local/lib/ruby/2.4.0/rexml/document.rb" line 45 in initialize
6  File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/xml/rexml.rb" line 7 in new
7  File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/xml/rexml.rb" line 7 in initialize
8  File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/xml.rb" line 59 in initialize
9  File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/api/errors.rb" line 49 in new
10 File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/api/errors.rb" line 49 in xml
11 File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/api/errors.rb" line 34 in description
12 File "/app/vendor/bundle/ruby/2.4.0/gems/recurly-2.16.2/lib/recurly/api/errors.rb" line 19 in to_s
...
```

As I say, as soon as we've logged the full message I'll let you know.

aaron-suarez:
Ok great, thank you! I may wait for your example so I can try it out. If it goes for very long time without you receiving an error, I'll play around and see if I can't find a condition where our API returns HTML instead of XML, and if that doesn't work out I may try to get it to blow up with a controlled experiment where I push HTML to the library.

But I'm going to wait on this to see if I get a real world example from you. Thanks for your patience.

bhelx:
@aaron-suarez It could be coming from cloudflare. Just a guess.

scottyp-env:
Hi @bhelx / @aaron-suarez, we've finally got a response that causes the error parser to choke. Looks like the cloudflare guess is a good one. Here's the error it failed on with identifying info removed.

```
Error while trying to parse Recurly error XML:
Missing end tag for 'meta' (got "head")
Line: 22
Position: 1072
Last 80 unconsumed characters:


Response Body:
<!DOCTYPE html>
<!--[if lt IE 7]> <html class="no-js ie6 oldie" lang="en-US"> <![endif]-->
<!--[if IE 7]>    <html class="no-js ie7 oldie" lang="en-US"> <![endif]-->
<!--[if IE 8]>    <html class="no-js ie8 oldie" lang="en-US"> <![endif]-->
<!--[if gt IE 8]><!--> <html class="no-js" lang="en-US"> <!--<![endif]-->
<head>
<meta http-equiv="refresh" content="0">

<title>envatoelements.recurly.com | 502: Bad gateway</title>
<meta charset="UTF-8" />
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
<meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1" />
<meta name="robots" content="noindex, nofollow" />
<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1" />
<link rel="stylesheet" id="cf_styles-css" href="/cdn-cgi/styles/cf.errors.css" type="text/css" media="screen,projection" />
<!--[if lt IE 9]><link rel="stylesheet" id='cf_styles-ie-css' href="/cdn-cgi/styles/cf.errors.ie.css" type="text/css" media="screen,projection" /><![endif]-->
<style type="text/css">body{margin:0;padding:0}</style>




</head>
<body>
<div id="cf-wrapper">

    

    <div id="cf-error-details" class="cf-error-details-wrapper">
        <div class="cf-wrapper cf-error-overview">
            <h1>
              
              <span class="cf-error-type">Error</span>
              <span class="cf-error-code">502</span>
              <small class="heading-ray-id">Ray ID: ... &bull; 2018-10-21 13:53:21 UTC</small>
            </h1>
            <h2 class="cf-subheadline">Bad gateway</h2>
        </div><!-- /.error-overview -->
        
        <div class="cf-section cf-highlight cf-status-display">
            <div class="cf-wrapper">
                <div class="cf-columns cols-3">
                  
<div id="cf-browser-status" class="cf-column cf-status-item cf-browser-status ">
  <div class="cf-icon-error-container">
    <i class="cf-icon cf-icon-browser"></i>
    <i class="cf-icon-status cf-icon-ok"></i>
  </div>
  <span class="cf-status-desc">You</span>
  <h3 class="cf-status-name">Browser</h3>
  <span class="cf-status-label">Working</span>
</div>

<div id="cf-cloudflare-status" class="cf-column cf-status-item cf-cloudflare-status ">
  <div class="cf-icon-error-container">
    <i class="cf-icon cf-icon-cloud"></i>
    <i class="cf-icon-status cf-icon-ok"></i>
  </div>
  <span class="cf-status-desc">Ashburn</span>
  <h3 class="cf-status-name">Cloudflare</h3>
  <span class="cf-status-label">Working</span>
</div>

<div id="cf-host-status" class="cf-column cf-status-item cf-host-status cf-error-source">
  <div class="cf-icon-error-container">
    <i class="cf-icon cf-icon-server"></i>
    <i class="cf-icon-status cf-icon-error"></i>
  </div>
  <span class="cf-status-desc">...</span>
  <h3 class="cf-status-name">Host</h3>
  <span class="cf-status-label">Error</span>
</div>

                </div>
              
            </div>
        </div><!-- /.status-display -->

        <div class="cf-section cf-wrapper">
            <div class="cf-columns two">
                <div class="cf-column">
                    <h2>What happened?</h2>
                    <p>The web server reported a bad gateway error.</p>
                </div>
              
                <div class="cf-column">
                    <h2>What can I do?</h2>
                    <p>Please try again in a few minutes.</p>
                </div>
            </div>
              
        </div><!-- /.section -->

        <div class="cf-error-footer cf-wrapper">
  <p>
    <span class="cf-footer-item">Cloudflare Ray ID: <strong>...</strong></span>
    <span class="cf-footer-separator">&bull;</span>
    <span class="cf-footer-item"><span>Your IP</span>: ...</span>
    <span class="cf-footer-separator">&bull;</span>
    <span class="cf-footer-item"><span>Performance &amp; security by</span> <a href="https://www.cloudflare.com/5xx-error-landing?utm_source=error_footer" id="brand_link" target="_blank">Cloudflare</a></span>
    
  </p>
</div><!-- /.error-footer -->


    </div><!-- /#cf-error-details -->
</div><!-- /#cf-wrapper -->
</body>
</html>
```

bhelx:
@scottyp-env thanks so much for that follow up! It's very helpful. I'd expect a `502` error to actually end up throwing a `GatewayError` on the client side: https://github.com/recurly/recurly-client-ruby/blob/9ef260900a7f6bb5da4df8f886aec0674fd9970f/lib/recurly/api/errors.rb#L202

Perhaps cloudflare is not setting the response code to 502.

scottyp-env:
(sorry for the repeat posts - I got bamboozled by the GitHub issue)

Yep, with our monkey patch in place we are seeing that `GatewayError` being logged now @bhelx. Without it though REXML's parsing error would raise another exception while trying to log the original, and so we didn't know we were getting a `GatewayError` because the original was never logged. This patch does the same as our monkey patch but without logging the response body of the non-XML error.

bhelx:
@scottyp-env That makes sense about the swallowing of exceptions. Still trying to think through a couple issues here. 

1. Ideally we won't swallow exceptions and return nil, though this might be a case in which we'd accept that.
2. For people using rexml, this could result in a weird error:
```
irb(main):003:0> def parse_xml
irb(main):004:1>   raise Error
irb(main):005:1> rescue  REXML::ParseException
irb(main):006:1>   puts "i will never be reached"
irb(main):007:1> end
=> :parse_xml
irb(main):008:0> parse_xml
Traceback (most recent call last):
        4: from /Users/ben/.rbenv/versions/2.5.1/bin/irb:11:in `<main>'
        3: from (irb):8
        2: from (irb):3:in `parse_xml'
        1: from (irb):5:in `rescue in parse_xml'
NameError (uninitialized constant REXML)
```

Right now only solution I can think that will work for everyone is to rescue all exceptions and return nil. Any ideas @aaron-suarez ?

aaron-suarez:
If I'm not mistaken, I believe that more users are using Nokogiri than REXML, so if Nokogiri users would see `NameError (uninitialized constant REXML)`, that would be extremely confusing. I agree that swallowing all errors and returning nil would be better than causing another unexplained error.

I'm going to continue to think on solutions that might be generic enough, but more specific than catching all exceptions and returning nil.

aaron-suarez:
Right now, when we initialize Nokogiri, we do so in a manner that successfully parses HTML as well as XML. We can change that behavior by changing this line https://github.com/recurly/recurly-client-ruby/blob/f64c918c40966e2be068e8d257e53c129849c811/lib/recurly/xml/nokogiri.rb#L5 to 

```ruby
        @root = Nokogiri::XML(xml) { |config| config.strict }.root
```

If this change is made, we can catch the two specific errors that you get when attempting to parse HTML as XML, and return `nil` in those situations. For any other errors, we can go ahead and just raise that same error. This way, if users are already catching certain errors from this method, they can continue to do so.

I think that stricter parsing rules for Nokogiri would be an improvement to the library even apart from the original issue that brought this up.

Anyway, then the rescue block could look something like this:
```ruby
rescue => e
  if defined?(Nokogiri) && e.is_a?(Nokogiri::XML::SyntaxError)
    nil
  elsif defined?(REXML) && e.is_a?(REXML::ParseException)
    nil
  else
    raise
  end
end```

This way we'd get similar behavior for users of REXML as well as Nokogiri, and we would hopefully limit the change to just the specific problem we're trying to solve.

**Comment by aaron-suarez** [comment]
I was doing some testing of the above. When we change Nokogiri to strict parsing, we'll also need to change this line:
https://github.com/recurly/recurly-client-ruby/blob/f64c918c40966e2be068e8d257e53c129849c811/spec/recurly/api/errors_spec.rb#L5
to some valid xml. Something as simple as
```ruby
let(:xml) { "<?xml version=\"1.0\"?>\n<root></root>" }
```
would do the trick. If we can do that, then we can take the `if` condition out of your test and it will work the same for Nokogiri as well as REXML.
```ruby
      describe "when response body is not valid XML" do
        let(:html) { "<html><body><hr></body></html>" }

        before { 3.times { response.expect :body, html } }

        it "must return nil" do
          error.send(:xml).must_equal nil
        end
      end
```

scottyp-env:
Thanks for the feedback @aaron-suarez. I've done a bit of thinking myself and have an idea built on top of yours.

If feels to me like checking for the the loaded XML library in the `Recurly::API::Errors` class is leaking implementation details a bit. How about if we catch the exceptions in the various adapters and raise a custom `Recurly::XML::ParseError` exception? Then we can just catch that in the `Recurly::API::Errors` class and can keep the `if defined?(...)` checks in one place. I'll tweak the PR with my idea so you can see what it looks like. What do you think?

aaron-suarez:
@bhelx Actually suggested to me on the side about putting in a comment that went something like this:
> \# TODO these should be caught in the xml adapters and re-raised as a single Recurly Xml parse exception

So it seems like you guys are on the same page with that, and I agree that it's a better idea than leaking the implementation details.

I wasn't going to ask you to do all that, but if you're willing, that contribution would be greatly appreciated as well!

PR #426, opened 2018-10-26
Remove Rails generator mention from README

lauramosher:
## What

In #388, the code for the rails config generator and railtie was removed. This updates the README to reflect the removal of the generator with a note to create ones own initializer for usage in Rails / maintaining old behavior.


bhelx:
@lauramosher thanks for catching this, looks like I completely forgot about it.

PR #428, opened 2018-11-03
Fix bug when company is not included in xml response

hasghari:
The 8908a771c commit broke the functionality for looking up an account and getting `company_name`.

This commit will fix that bug while preserving the intended functionality.

### DIFF
--- a/lib/recurly/account.rb
+++ b/lib/recurly/account.rb
@@ -85,7 +85,10 @@ module Recurly
       preferred_locale
     )
     alias to_param account_code
-    alias company_name company
+
+    def company_name
+      super || company
+    end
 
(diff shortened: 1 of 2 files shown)


hasghari:
@aaron-suarez Thanks for merging this. Is there going to be a new release soon?

aaron-suarez:
The next scheduled release is for the API v2.17 changes on December 11. If this is a blocker for you, I can see about doing a patch release sooner than that.

hasghari:
I see. A patch release would be nice since it fixes a bug. This is preventing us from upgrading our gem version to "2.17" right now.

hasghari:
wonderful, thank you.

Issue #431, opened 2018-11-16
When an adjustment type is nil?


spyrbri:
In the Recurly docs says:
`The type of adjustment to return: charge or credit.`

https://dev.recurly.com/docs/adjustment-object

In the gem there is a case where type is nil. 

https://github.com/recurly/recurly-client-ruby/blob/1446464b6ae1e4537fee68f657c4a18b9336043a/lib/recurly/adjustment.rb#L60

 When does this happen?

aaron-suarez:
@spyrbri That is an excellent question. The easiest demo is if you are creating a new Adjustment object (I'm not sure why you might do this in production, but technically the library allows this):

```ruby
[1] pry(main)> adjustment = Recurly::Adjustment.new
=> #<Recurly::Adjustment accounting_code: nil, created_at: nil, credit_reason_code: nil, currency: "USD", description: nil, discount_in_cents: nil, end_date: nil, origin: nil, original_adjustment_uuid: nil, product_code: nil, proration_rate: nil, quantity: nil, quantity_remaining: nil, revenue_schedule_type: nil, shipping_address_id: nil, start_date: nil, state: nil, tax_code: nil, tax_details: nil, tax_exempt: nil, tax_in_cents: nil, tax_rates: nil, tax_region: nil, tax_type: nil, tax_types: nil, total_in_cents: nil, unit_amount_in_cents: nil, updated_at: nil, uuid: nil>
[2] pry(main)> adjustment.type
=> nil
```

That being said, is the real question whether the API will ever return an adjustment that has a nil type? I'm not sure the answer to that, but in 5 minutes of testing various ways to create an adjustment, I was unable to create on my sandbox site an adjustment that was not either a `charge` or a `credit`. If you're concerned about this, I can dig a little deeper, but I believe that's there because it's technically possible, not because it's supposed to ever be `nil`.

aaron-suarez:
This is going to take a little investigating for me to reproduce. If either of you can provide example XML that causes this (and @spyrbri if you can please indicate whether this happens with Webhooks or via an API request, that would help as well).

If there's a problem with an API request, please indicate which method on what type of object causes this behavior. e.g., `adjustment.find 'something'`

I'll do some testing on my own based on the information I have so far and see if we can get to the bottom of this.

aaron-suarez:
Ok, just so we're on the same page:

When an adjustment comes in from the API, its root element looks like this, with the attributes `href` and `type`:
```xml
[...]

PR #436, opened 2018-11-27
Fix missing type instance variable and add specs

spyrbri:
We are missing several adjustment types after caching our recurly resources.

https://github.com/recurly/recurly-client-ruby/issues/431

That's because in https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L793
the objects are saved as Marshal dumps.

After the marshal transformation we get no type attribute because of the overriden marshal methods that lack the type instance variable.


aaron-suarez:
@spyrbri First of all, good catch! I learned a lot about Marshalling today trying to understand this PR and its implications.

Second, after much reading, I had this uneasy feeling regarding making this change directly on the Resource class. I'm not sure if this is actually a problem or not, so let's talk through it and see if we can come to an agreement about this.

There are 4 classes in the library that extend Resource that have an `@type` instance variable set with the `type` attribute from the root tag of the XML object:
`transaction`
`credit_payment`
`adjustment`
`billing_info`

As you pointed out in the issue, `invoice` also has a type that it gets from a `<type>` tag.

My concern is regarding other classes that extend Resource that do not have a type. To try this out, I went ahead and copied your spec to another Resource that does not have a type attribute, in this case `coupon`:

```ruby
  describe '#marshal_dump' do
    it 'must return the same instance variables' do
      coupon_from_dump = Marshal.load(Marshal.dump(coupon))

      coupon.instance_variables.must_equal coupon_from_dump.instance_variables
    end
  end
```

Note that the stubbed API request is set in a before action on this spec, so all I needed to replicate was the loading/dumping and the assertion.

This spec fails as so:

```
  1) Failure:
Recurly::Coupon::#marshal_dump#test_0001_must return the same instance variables [/Users/asuarez/code/v2/clients/ruby/spec/recurly/coupon_spec.rb:58]:
--- expected
+++ actual
@@ -1 +1 @@
-[:@href, :@uri, :@destroyed, :@new_record, :@attributes, :@changed_attributes, :@links, :@etag, :@response, :@previous_changes, :@type]
+[:@href, :@uri, :@destroyed, :@new_record, :@attributes, :@changed_attributes, :@links, :@etag, :@response, :@previous_changes]
```

I'm not sure if adding the `@type` attribute to `marshal_load` and `marshal_dump` would actually be a problem for serializing/deserializing these other resources that don't have a `@type` instance variable. Like I said, it was more of an uneasy feeling than a hard "no".

Do you think that there would be an easy way to limit this change to just the Resources that indeed have a `@type` ivar? If not, can you think of a way that we can verify that this won't cause any problems with any type-less Resources in anyone's implementation of the library?

lib/recurly/resource.rb
```diff
@@ -1000,6 +1000,7 @@ def marshal_dump
         @destroyed,
         @uri,
         @href,
+        @type,
```
aaron-suarez: To prevent problems with upgrading to this change, we need to move the declaration for `@type` to the end of the array here.

lib/recurly/resource.rb
```diff
@@ -1014,6 +1015,7 @@ def marshal_load(serialization)
...
```
aaron-suarez: Similarly, `@type` needs to be moved after `links` in this method. The reason for this is that if a user marshalled something with an old version of the library and then upgrades to a version that includes this change, it will assign the wrong data to the attribute. If we put it at the end of the array, it will just be nil if it's missing from the array altogether ensuring compatibility between versions of the library.

It won't look as pretty in the code, but it will be safer.

spyrbri:
@aaron-suarez I understand your concern. You are right, I forgot about all the other resources. We could override the marshal methods of the Resource class into the subclasses to be more safe. 
Do you have any other idea?

aaron-suarez:
I was thinking something like that. For each of the 5 classes that need a `@type`, we could override `marshal_load` and `marshal_dump` to include it... I just hope that if a `type` attribute is added by the API to any other classes, we ensure that we override it there as well.

If we go that route, it would be nice to add a comment linking to this PR so future maintainers could easily see why we did this in the classes rather than in `Resource` directly.

If you want to go ahead and update the PR as discussed, I have no problem merging it. Please still put the `@type` attribute at the end of the list when you make your changes.

Thanks!

spyrbri:
> I just hope that if a type attribute is added by the API to any other classes, we ensure that we override it there as well.

We could add a spec in the resource and if the subclass has a type instance variable should also have it in the marshal methods.

> If we go that route, it would be nice to add a comment linking to this PR so future maintainers could easily see why we did this in the classes rather than in Resource directly.

Sure

spyrbri:
Updated, let me know if it looks ok now. I will squash the first commit when the PR is ready to be merged.

spec/fixtures/transactions/show-200.xml
```diff
@@ -2,7 +2,7 @@ HTTP/1.1 200 OK
 Content-Type: application/xml; charset=utf-8
 
 <?xml version="1.0" encoding="UTF-8"?>
-<transaction href="https://api.recurly.com/v2/transactions/abcdef1234567890">
+<transaction href="https://api.recurly.com/v2/transactions/abcdef1234567890" type="credit_card">
```
aaron-suarez: Nit: `type="credit"` is what would come from the API. Doesn't really make a difference for the spec though

aaron-suarez:
In https://github.com/recurly/recurly-client-ruby/issues/431#issuecomment-440935348 you expressed a concern about the Invoices class as well. Do you want to go ahead and add a `<type>charge</type>` tag to `spec/fixtures/invoices/show-200.xml` and add the following spec to `invoices.rb`?

```ruby
    describe 'marshal methods' do
      it 'must return the same instance variables' do
        stub_api_request :get, 'invoices/show-invoice', 'invoices/show-200'
        invoice = Invoice.find 'show-invoice'
        invoice_from_dump = Marshal.load(Marshal.dump(invoice))
        invoice.instance_variables.must_equal invoice_from_dump.instance_variables
      end
       it 'must return the same values' do
        stub_api_request :get, 'invoices/show-invoice', 'invoices/show-200'
        invoice = Invoice.find 'show-invoice'
        invoice_from_dump = Marshal.load(Marshal.dump(invoice))
        invoice.type.must_equal invoice_from_dump.type
      end
    end
```

Just to make sure we have all our bases covered.

spec/fixtures/transactions/show-200.xml
spyrbri: I added it by checking https://dev.recurly.com/docs/transaction-object
which has `credit_card` type, right?

spyrbri:
> In [#431 (comment)](https://github.com/recurly/recurly-client-ruby/issues/431#issuecomment-440935348) you expressed a concern about the Invoices class as well. Do you want to go ahead and add a `<type>charge</type>` tag to `spec/fixtures/invoices/show-200.xml` and add the following spec to `invoices.rb`?
> 
> ```ruby
>     describe 'marshal methods' do
>       it 'must return the same instance variables' do
>         stub_api_request :get, 'invoices/show-invoice', 'invoices/show-200'
>         invoice = Invoice.find 'show-invoice'
>         invoice_from_dump = Marshal.load(Marshal.dump(invoice))
>         invoice.instance_variables.must_equal invoice_from_dump.instance_variables
>       end
>        it 'must return the same values' do
>         stub_api_request :get, 'invoices/show-invoice', 'invoices/show-200'
>         invoice = Invoice.find 'show-invoice'
>         invoice_from_dump = Marshal.load(Marshal.dump(invoice))
>         invoice.type.must_equal invoice_from_dump.type
>       end
>     end
> ```
> 
> Just to make sure we have all our bases covered.

Ok, I didn't do that because we have seen from our production logs that we don't receive the type element in the XMLs at all, but better safe than sorry.

lib/recurly/transaction.rb
```diff
@@ -125,5 +125,36 @@ def inspect
       end
       super attributes
     end
+
+    # https://github.com/recurly/recurly-client-ruby/pull/436
+    def marshal_dump
+      [
+        @attributes.reject { |k, v| v.is_a?(Proc) },
```
bhelx: Could we change this definition to use the code from the resource class if possible?

```ruby
def marshal_dump
  super << @type
end
```

lib/recurly/transaction.rb
```diff
@@ -125,5 +125,36 @@ def inspect
...
+        changed_attributes,
+        previous_changes,
+        response,
+        etag,
+        links,
+        @type
+      ]
+    end
+
+    def marshal_load(serialization)
```
bhelx: I think it should be safe to add `@type` to the list in `Resource`'s definition. It will just be nil for any object serialized with a previous version.

bhelx: So we probably don't need this method here.

lib/recurly/transaction.rb
bhelx: @aaron-suarez do we need this implementation at all?

spyrbri: I thought that too but I felt it is clearer to have those methods displayed together on the class.

lib/recurly/transaction.rb
spyrbri: Do you think we will get any benefit from that except for less code?

spyrbri:
I will go on and update as per @bhelx comments. If we change our mind we just remove the commit.

spec/fixtures/transactions/show-200.xml
aaron-suarez: Oh, I was basing it off of this comment
https://github.com/recurly/recurly-client-ruby/blob/master/lib/recurly/transaction.rb#L82

I'm not sure whether the docs are wrong or this comment is wrong, but I would tend to think the docs are probably right... I don't know if I care enough to look into it right this second, leave it like the docs and I'll update that comment in the code eventually if I find out an answer to that

lib/recurly/transaction.rb
spyrbri: Ok, I get it. The only fear is that this extra type instance variable after marshaling could cause some other weird issue.

lib/recurly/transaction.rb
bhelx: My assumption was that we don't need to override these methods at all. We should be able to add `@type` to both marshal_dump and marshal_load. That would just mean that every `Resource` would have a `@type` instance variable but only the ones that use that xml attribute will have the `attr_reader :type` so it won't be exposed in resources where it's always nil (and it would be nil since extra variables in a destructuring statement are set to nil). We'd probably want to change places that look for the @type to be defined to instead ask if the object `respond_to?(:type)`. Example:
https://github.com/recurly/recurly-client-ruby/blob/d5d5c7bac5ee5308c3b2f38aa1d758ca9bf30acb/lib/recurly/resource.rb#L1027

I'm okay with the current path the code is taking too. I haven't tested this and there might be things I'm missing. I'm just providing some feedback but the final decision is for @aaron-suarez . Let's do whatever works 👍

aaron-suarez:
I'm sorry about all the back and forth on this.

Right now, I don't have a specific bug that I know for a fact adding `@type` to the resource marshal_load/dump methods would cause for other classes such as `Coupon` that don't have a type. It's very possible it wouldn't cause any problems at all, as long as we put the `@type` at the bottom of the marshal_load/dump lists.

I think it's fairly safe as is, but harder to maintain in the future.

@spyrbri What do you think about this? Would it be better to add a `@type` variable to the Resource class, and just define it as `nil` by default? Then every Resource would have a type, and if it happens to come with a `type` attribute in the XML, it would be set, otherwise, it would be nil. Then the only change that would have to be made is in the Resource class and we can get rid of all the changes in the subclasses.

The only thing to think about would be here:

https://github.com/recurly/recurly-client-ruby/blob/1446464b6ae1e4537fee68f657c4a18b9336043a/lib/recurly/resource.rb#L1027

We'd probably be able to change that from `instance_variable_defined?` to `respond_to?(:type)`...

I wish I had thought about that before having you implement it in all the subclasses, and I understand if you're kind of over this PR and just want to get it merged. I would be fine merging as is. But if you agree that the above would be a more elegant solution, we could do that.

Oh, and another benefit to that is if the API changes such that for some reason a type is added to `Coupon` (or any other Resource), we wouldn't have to change the code in the subclass at all, it would work out of the box and not have any marshalling problems either.

aaron-suarez:
@bhelx I'm going to leave this unmerged just to make sure you're ok with it too. It looks good to me so long as the specs pass

spyrbri:
I removed the transaction spec by mistake. I added it back.

PR #438, opened 2018-12-06
Add OpenSSL version to user_agent

aaron-suarez:
Adds OpenSSL version to the user agent.

### DIFF
--- a/lib/recurly/api.rb
+++ b/lib/recurly/api.rb
@@ -85,7 +85,7 @@ module Recurly
 
       # @return [String]
       def user_agent
-        "Recurly/#{Version}; #{RUBY_DESCRIPTION}"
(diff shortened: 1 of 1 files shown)


lib/recurly/api.rb
```diff
@@ -85,7 +85,7 @@ def validate_uri!(uri)
 
       # @return [String]
       def user_agent
-        "Recurly/#{Version}; #{RUBY_DESCRIPTION}"
+        "Recurly/#{Version}; #{RUBY_DESCRIPTION} #{OpenSSL::OPENSSL_LIBRARY_VERSION}"
```
bhelx: Should we consider putting a delimiter in here? perhaps another semicolon.

bhelx:
👍

PR #442, opened 2018-12-18
Cleanup insecure nokogiri or ruby msg

bhelx:
This slightly better reflects the case of having a supported version of ruby with an old version of nokogiri.

### DIFF
--- a/lib/recurly/xml.rb
+++ b/lib/recurly/xml.rb
@@ -89,9 +89,8 @@ if defined? Nokogiri
   insecure_noko_msg = <<-MSG
 
     You are attempting to use an insecure version of
-    nokogiri on an insecure version of ruby. Please see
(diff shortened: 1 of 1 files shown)


lib/recurly/xml.rb
```diff
@@ -89,9 +89,9 @@ def to_s
   insecure_noko_msg = <<-MSG
 
     You are attempting to use an insecure version of
-    nokogiri on an insecure version of ruby. Please see
+    nokogiri or an insecure version of ruby. Please see
```
aaron-suarez: Should we just consider showing a completely different error for nokogiri vs ruby? Maybe it doesn't matter that much, but at the same time, we're checking for all these different scenarios, we might as well tell the user which one needs to be upgraded to pass our checks.

PR #444, opened 2019-01-03
Add support for account hierarchy

aaron-suarez:
This PR adds support for account hierarchy.

```ruby
account = Recurly::Account.create(
  :account_code => 'example-123',
  :parent_account_code => 'example-124'
)
```

### DIFF
--- a/lib/recurly/account.rb
+++ b/lib/recurly/account.rb
@@ -42,6 +42,12 @@ module Recurly
     # @return [AccountBalance, nil]
     has_one :account_balance, readonly: true
 
+    # @return [Account, nil]
+    belongs_to :parent_account, class_name: :Account
+
+    # @return [Pager<Account>, []] A pager that yields Account for persisted
+    has_many :child_accounts, class_name: :Account
+
     # @return [Pager<CreditPayment>, []]
     has_many :credit_payments, class_name: :CreditPayment, readonly: true
 
@@ -57,6 +63,7 @@ module Recurly
 
     define_attribute_methods %w(
       account_code
+      parent_account_code
       state
       username
       email
--- a/spec/fixtures/accounts/hierarchy/create-201.xml
+++ b/spec/fixtures/accounts/hierarchy/create-201.xml
@@ -1,0 +1,35 @@
+HTTP/1.1 201 Created
+Content-Type: application/xml; charset=utf-8
+Location: https://api.recurly.com/v2/accounts/abcdef1234567890.xml
+
+<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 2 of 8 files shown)


aaron-suarez:
I'm not going to include a spec for associating a parent account via a subscription or purchase because that's more of a behavior of the API rather than the code (plus, the response for these doesn't actually include a reference to the parent or child accounts). But here are the scripts just in case it's relevant to someone:

Associate a parent via create subscription:
```ruby
# Note: this assumes example-124 (the parent account) already exists
# Assumes plan code 'silver' is a valid plan

  subscription = Recurly::Subscription.create!(
    # subscription, account data
    :plan_code => 'silver',
    :currency  => 'USD',
    :account   => {
      :account_code => 'example-123',
      :parent_account_code => 'example-124'
    }
  )
```
Remove parental association via create subscription:
```ruby
# Note: this assumes that a parent account was associated with account 'example-123'
# Assumes plan code 'silver' is a valid plan

  subscription = Recurly::Subscription.create!(
    # subscription, account data
    :plan_code => 'silver',
    :currency  => 'USD',
    :account   => {
      :account_code => 'example-123',
      :parent_account_code => ''
    }
  )
```
Associate a parent via Purchase:
```ruby
# Assumes 'example-123' and 'example-124' exist
# Assumes plan code 'silver' is a valid plan

purchase = Recurly::Purchase.new({
  currency: 'USD',
  collection_method: :automatic,
  account: {
    account_code: 'example-123',
    parent_account_code: 'example-124',
  },
  subscriptions: [
    {
      plan_code: "silver"
    }
  ]
})

collection = Recurly::Purchase.invoice!(purchase)
```
Remove association of parent account via Purchase:
```ruby
# Assumes 'example-123' and has a parent account
# Assumes plan code 'silver' is a valid plan

purchase = Recurly::Purchase.new({
  currency: 'USD',
  collection_method: :automatic,
  account: {
    account_code: 'example-123',
    parent_account_code: '',
  },
  subscriptions: [
    {
      plan_code: "silver"
    }
  ]
})

collection = Recurly::Purchase.invoice!(purchase)
```

lib/recurly/account.rb
```diff
@@ -42,6 +42,12 @@ class Account < Resource
     # @return [AccountBalance, nil]
     has_one :account_balance, readonly: true
 
+    # @return [Account, nil]
+    belongs_to :parent_account, class_name: "Account"
```
bhelx: We have a convention of using symbols for the account class name. Is there a reason we are using string here?

aaron-suarez: No, I just didn't look at other uses of `belongs_to`. I'll change that to a symbol.

PR #449, opened 2019-02-11
Sonarcloud Test

unbounceRabbit:


lib/recurly.rb
```diff
@@ -104,6 +104,13 @@ def log level, message
       logger.send(level, name) { message }
     end
 
+    # Returns a boolean value to tell recurly whether or not to log the response in
+    # the rails log.  We only do this if the environment is production because the
+    # response and response.body are quite helpful when it comes to debugging
```
aaron-suarez: Friendly reminder to be very careful not to log any of your customer's PII in production, as it is not only a security risk, but it can also increase your PCI scope (meaning it could result in serious fines if this data is mishandled).

PR #450, opened 2019-02-11
Add support for Account Acquisition

aaron-suarez:
This PR adds support for [Account Acquisition](https://docs.recurly.com/docs/account-acquisition-data).

Create script:
```ruby
acct = Recurly::Account.find 'x'

acct.account_acquisition = {
  cost_in_cents: 300,
  currency: "USD",
  channel: "marketing_content",
  subchannel: "The coolest content we have made them sign up!",
  campaign: "mailchimp67a904de95.0914d8f4b4"
}

acct.save!
```
Lookup an account acquisition:
```ruby
acct = Recurly::Account.find 'x'
acquisition = acct.account_acquisition
puts acquisition
```
Update:
```ruby
acct = Recurly::Account.find 'x'

acct.account_acquisition.cost_in_cents = 299
acct.account_acquisition.subchannel = "The 2nd coolest content we have made them sign up!"

acct.account_acquisition.save!
```
Delete:
```ruby
acct = Recurly::Account.find 'x'
if acquisition = acct.account_acquisition
  acquisition.destroy
end
```

### DIFF
--- a/lib/recurly/account.rb
+++ b/lib/recurly/account.rb
@@ -48,6 +48,9 @@ module Recurly
     # @return [[CustomField], []]
     has_many :custom_fields, class_name: :CustomField, readonly: false
 
+    # @return [AccountAcquisition, nil]
+    has_one :account_acquisition, class_name: :AccountAcquisition, readonly: false
+
     # Get's the first redemption given a coupon code
     # @deprecated Use #{redemptions} instead
     # @param coupon_code [String] The coupon code for the redemption
--- a/spec/recurly/account_spec.rb
+++ b/spec/recurly/account_spec.rb
@@ -341,4 +341,19 @@ XML
       account.company_name.must_equal 'My Company Inc.'
     end
   end
+
+  describe "account acquisition" do
+    let(:account) {
+      stub_api_request :get, 'accounts/abcdef1234567890', 'accounts/show-200'
+      stub_api_request :get, 'accounts/abcdef1234567890/acquisition', 'account_acquisition/show-200'
+      Account.find 'abcdef1234567890'
+    }
+
+    it 'is able to retrieve and parse acquisition' do
+      acquisition = account.account_acquisition
+      acquisition.cost_in_cents.must_equal 299
+      acquisition.channel.must_equal "blog"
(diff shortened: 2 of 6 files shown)


spec/recurly/account_spec.rb
```diff
@@ -341,4 +341,19 @@
...
   end
+
+  describe "account acquisition" do
+    let(:account) {
+      stub_api_request :get, 'accounts/abcdef1234567890', 'accounts/show-200'
+      stub_api_request :get, 'accounts/abcdef1234567890/acquisition', 'account_acquisition/show-200'
+      Account.find 'abcdef1234567890'
+    }
+
+    it 'is able to retrieve and parse acquisition' do
```
aaron-suarez: I only did this spec because `destroy` seems to be covered in the resource spec, and `save` is already covered above. I can add a method to the `save` describe block if requested

lib/recurly/account.rb
```diff
@@ -48,6 +48,9 @@ class Account < Resource
     # @return [[CustomField], []]
     has_many :custom_fields, class_name: :CustomField, readonly: false
 
+    # @return [AccountAcquisition, nil]
+    has_one :account_acquisition, readonly: false
```
bhelx: I generally like to specify the `class_name` here even if it doesn't need it. This library has some quirks when it tries to guess the class name. Is there a reason you didn't provide it?

lib/recurly/account.rb
```diff
@@ -57,6 +60,7 @@ def redemption(coupon_code)
 
     define_attribute_methods %w(
       account_code
+      account_acquisition
```
bhelx: `has_one` defines some methods on this class and this redefines them. Doing both could cause some weirdness. standard practice is to leave this out.

lib/recurly/account.rb
aaron-suarez: I didn't provide it because it appeared to work without it. There's a balance between the fact that less code is often less complexity, and explicit should be preferred over implicit. So I'll add it back in based on the latter and the fact that I don't know all the quirks you're referring to.

PR #455, opened 2019-02-26
Fix NoMethodError thrown by Resource.invalid!

### LINKED ISSUE #400: Invalid subscription add_on leads to unhandled exceptions
When making a `Recurly::Purchase.invoice!(purchase)` with a purchase that includes a subscription with invalid add_ons (eg: non-existent add_on_code) the following unhandled exception occurs

```ruby
NoMethodError: undefined method `invalid!' for #<Recurly::Subscription::AddOns:0x00007f89ffe88510>
```

I think this is because the recursive `invalid!` error handling expects everything along the way to be a `Resource` but `Recurly::Subscription::AddOns` are not.

A cursory examination of the API response suggests the solution may be non trivial

```xml
<error field="purchase.subscriptions[0].subscription_add_ons.add_on" symbol="">is invalid</error>
```

I would have expected the API to specify _which_ was the offending subscription_add_on

aaron-suarez:
Fixes #400 

A `NoMethodError` was being thrown when the `invalid!` method was being called on children of a Resource that are not indeed Resources themselves. There were two examples of this found in the wild:
1) When a purchase was made with a subscription that had an invalid subscription add on, the subscription add on object is a `Pager`, not a `Resource`, and therefore had no `invalid!` method.

2) When a plan was updated with an invalid currency, the `plan.unit_amount_in_cents` is a `Recurly::Money` rather than a `Resource`, and therefore does not have an `invalid!` method.

The library previously assumed that all children of `Resource`s are also `Resource`s. This PR fixes a particular instance of this assumption.

The included specs are similar to the actual reports from #400. The code change was also tested against the actual API and expected behavior was exhibited. Below, please see my test scripts for manual testing:

Purchase with bad add on:
```ruby
# Assumes a plan called "gold" is set up on the account
require_relative './stub'

require 'securerandom'

purchase = Recurly::Purchase.new({
  currency: 'USD',
  collection_method: :automatic,
  po_number: '98675',
  account: {
    account_code: SecureRandom.uuid,
    billing_info: {
      first_name: 'Aaron',
      last_name: 'Du Monde',
      address1: '400 Alabama St',
      city: 'San Francisco',
      state: 'CA',
      zip: '94110',
      country: 'US',
      number: '4111-1111-1111-1111',
      month: 12,
      year: 2019,
    },
  },
  subscriptions: [
    {
      plan_code: "gold",
      subscription_add_ons: [
        add_on_code: 'invalid_code',
        unit_amount_in_cents: 200
      ]
    }
  ]
})

begin
  collection = Recurly::Purchase.invoice!(purchase)
  puts collection.inspect
rescue Recurly::Resource::Invalid => e
  puts e.inspect
  # Invalid data
rescue Recurly::Transaction::Error => e
  puts e.inspect
  # Transaction error
  # e.transaction
  # e.transaction_error_code
end
```
Plan with currency that is not configured on the site:
```ruby
require_relative './stub'
# Assumes that "silver" is a valid plan on the site
# Assumes that currency "BRL" is not configured on the site

plan_code = "silver"
plan = Recurly::Plan.find(plan_code)
plan.unit_amount_in_cents['BRL'] = 329_00
plan.save!
```

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -1068,7 +1068,7 @@ module Recurly
       else
         child, k, v = attribute_path.shift.scan(/[^\[\]=]+/)
         if c = k ? self[child].find { |d| d[k] == v } : self[child]
-          c.invalid! attribute_path, error
+          c.invalid! attribute_path, error if c.methods.include? :invalid!
           e = errors[child] << 'is invalid' and e.uniq!
         end
       end
--- a/spec/fixtures/plans/show-422.xml
+++ b/spec/fixtures/plans/show-422.xml
@@ -1,0 +1,8 @@
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/xml; charset=utf-8
+
+<?xml version="1.0" encoding="UTF-8"?>
+<errors>
+  <error field="plan.unit_amount_in_cents.BRL" symbol="inclusion">is not accepted by the site</error>
+  <error field="plan.setup_fee_in_cents.BRL" symbol="inclusion">is not accepted by the site</error>
+</errors>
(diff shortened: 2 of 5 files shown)


bhelx:
Sorry for the delay on this everyone! Thanks @aaron-suarez

PR #468, opened 2019-04-04
Make the set_options method public

aaron-suarez:
I got the `collect_invoice.rb` script working, but it requires a certain header added to the request. This is a first iteration ("just get it working" kind of thing), so I'm open to other strategies. However, if we want that script to pass, we'll need that header so we need a way to set custom headers on the client somehow. This was the way that seemed built in, so it was minimal work to just make it public.

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -69,6 +69,15 @@ module Recurly
       end
     end
 
+    def set_options(options)
+      @log_level = options[:log_level] || Logger::WARN
+      @logger = Logger.new(STDOUT)
+      @logger.level = @log_level
+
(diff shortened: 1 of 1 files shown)


bhelx:
What does it look like to use this? This wasn't really meant to be exposed.

aaron-suarez:
In the case of `collect_invoice`, I need to add the skip auth header, so I use set_options like so:
```ruby
@client.set_options(headers: { "Recurly-Skip-Authorization" => "true" })
```

aaron-suarez:
Since it wasn't meant to be exposed, we're going to solve this a different way

PR #469, opened 2019-04-17
Add strict mode for json deserializer

bhelx:
Adds a "strict" mode which makes the client more conservative about parsing / serializing. If it finds something that is off, it will throw an exception. This will be mostly useful for smoke testing.

### DIFF
--- a/lib/recurly/schema/json_deserializer.rb
+++ b/lib/recurly/schema/json_deserializer.rb
@@ -44,8 +44,9 @@ module Recurly
 
             resource.send(writer, val)
           else
-            # TODO ignoring these missing fields for now
-            puts "Missing #{attr_name}"
+            if Recurly::STRICT_MODE
+              raise ArgumentError, "#{resource.class.name} encountered json attribute #{attr_name.inspect}: #{val.inspect} but it's unknown to it's schema"
+            end
           end
         end
         resource
--- a/lib/recurly.rb
+++ b/lib/recurly.rb
@@ -13,4 +13,8 @@ require "recurly/errors"
 require "recurly/client"
 
 module Recurly
+  STRICT_MODE = !ENV['RECURLY_STRICT_MODE'].nil?
+  if STRICT_MODE
+    puts "[Recurly] [WARNING] STRICT_MODE enabled. This should only be used for testing purposes."
+  end
 end


lib/recurly/schema/json_deserializer.rb
```diff
@@ -44,8 +44,9 @@ def from_json(attributes = {})
 
             resource.send(writer, val)
           else
-            # TODO ignoring these missing fields for now
-            puts "Missing #{attr_name}"
+            if Recurly::STRICT_MODE
+              raise ArgumentError, "#{resource.class.name} encountered json attribute: #{{ attr_name => val.inspect }.inspect} but it's unknown to it's schema"
```
aaron-suarez: So this would be raised in a situation where the API was updated with an additional attribute but the client never got that update and so it fails because there's an extra attribute that shouldn't be there?

aaron-suarez: We should also include the name of the attribute `attr_name` in the log. Right now it prints something like this
```
Recurly::Resources::CouponRedemption encountered json attribute: {"account"=>"{\"id\"=>\"kp5fz0m8uy3y\", \"object\"=>\"account\", \"code\"=>\"bb9639b8e6a@example.com\", \"email\"=>nil, \"first_name\"=>nil, \"last_name\"=>nil, \"company\"=>nil, \"parent_account_id\"=>nil, \"bill_to\"=>\"self\"}"} but it's unknown to it's schema (ArgumentError)
```
And even though it's clear what the value of the JSON attribute is, it's name is not logged. Suggest we change it to something like this:
```ruby
raise ArgumentError, "#{resource.class.name} encountered json attribute '#{attr_name}': #{{ attr_name => val.inspect }.inspect} but it's unknown to it's schema"
```
which would come out like this:
```
Recurly::Resources::CouponRedemption encountered json attribute 'account': {"account"=>"{\"id\"=>\"kp5hfj1bsmuo\", \"object\"=>\"account\", \"code\"=>\"69b2e2cfbee@example.com\", \"email\"=>nil, \"first_name\"=>nil, \"last_name\"=>nil, \"company\"=>nil, \"parent_account_id\"=>nil, \"bill_to\"=>\"self\"}"} but it's unknown to it's schema (ArgumentError)
```

bhelx: Yes, the grand idea is for it to explode if there is a mismatch b/w it's internal schema and what it's seeing in the responses.

PR #490, opened 2019-08-09
CurrencyArrays

bhelx:
Pricings in the API are often presented as an array of objects. As a convenience, this library will turn these into
a CurrencyArray. This class is a subclass of Array that allows you to index by currency. Example, consider this plan with
a currencies attribute:

```ruby
plan
#=> #<Recurly::Resources::Plan:0x0000556b25c2ca28
#=> @attributes=
#=>  {
#=>  #...
#=>  :currencies=>[#<Recurly::Requests::PlanPricing:0x0000556b25c21538 @attributes={:currency=>"USD", :setup_fee=>1000.0, :unit_amount=>234.0}>],
#=>  #...
#=>  }>
```

You can now fetch a PlanPricing by currency rather than searching the array or knowing the index:

```ruby
plan.currencies[:usd].unit_amount # 234.0
plan.currencies[:USD].unit_amount # 234.0
plan.currencies["USD"].unit_amount # 234.0
# positional index still works
plan.currencies[0].unit_amount # 234.0
```

### DIFF
--- a/lib/recurly/schema.rb
+++ b/lib/recurly/schema.rb
@@ -53,21 +53,6 @@ module Recurly
       end
     end
 
-    # Describes a list attribute type
-    class List
-      # The type of the elements of the list
-      # @return [Symbol]
-      attr_accessor :item_type
-
-      def initialize(item_type)
-        @item_type = item_type
-      end
-
-      def to_s
-        "List<#{item_type}>"
-      end
-    end
-
     # Describes and attribute for a schema.
     class Attribute
       # The name of the attribute.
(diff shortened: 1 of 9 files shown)


lib/recurly/schema.rb
```diff
@@ -53,21 +53,6 @@ def self.get_recurly_class(type)
       end
     end
 
-    # Describes a list attribute type
-    class List
```
bhelx: Appears this was never used.

lib/recurly/schema/json_deserializer.rb
```diff
@@ -27,13 +27,20 @@ def from_json(attributes = {})
...
+                    items = val.map do |e|
                       if e.is_a?(Hash) && !schema_attr.is_primitive?
                         schema_attr.recurly_class.from_json(e)
                       else
                         e
                       end
                     end
+                    # If the item type has a "currency" key, let's assume we can
+                    # safely index by currency and return a CurrencyArray
+                    if !schema_attr.is_primitive? && schema_attr.recurly_class.schema.get_attribute(:currency)
```
bhelx: For now, we are just looking to see if the item's type has a `currency` in the schema. In the future, I may make this a first class citizen in the recurly api spec.

PR #510, opened 2019-09-19
Refactor internal schema representation

bhelx:
This restructures the schema representation to a similar style I took with the node client: https://github.com/recurly/recurly-client-node/blob/e4f03fc1e831aeb773a574dd9823255ab746ee93/lib/recurly/schemas.js

It makes the code cleaner and easier to test, and also means a lot fewer conditionals and checks at runtime so it's more performant as well.

Here's before and after on the JSON parsing / casting of a response:

```
       user     system      total        real
  0.210000   0.000000   0.210000 (  0.212076)
```
```
       user     system      total        real
  0.130000   0.000000   0.130000 (  0.133469)
```

### DIFF
--- a/benchmark.rb
+++ b/benchmark.rb
@@ -1,0 +1,16 @@
+require "rubygems"
+require "bundler/setup"
+require "recurly"
+require "benchmark"
+
+N = 1_000
+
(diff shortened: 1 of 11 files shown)


benchmark.rb
```diff
@@ -0,0 +1,16 @@
...
+require "recurly"
+require "benchmark"
+
+N = 1_000
+
+account_body = "{\"id\":\"ljvmmbjchtgs\",\"object\":\"account\",\"code\":\"9ebd49f7288@example.com\",\"parent_account_id\":null,\"bill_to\":\"self\",\"state\":\"active\",\"username\":null,\"email\":null,\"cc_emails\":null,\"preferred_locale\":null,\"first_name\":\"Benjamin\",\"last_name\":\"Du Monde\",\"company\":null,\"vat_number\":null,\"tax_exempt\":false,\"exemption_certificate\":null,\"address\":null,\"billing_info\":null,\"shipping_addresses\":[{\"object\":\"shipping_address\",\"first_name\":\"Benjamin\",\"last_name\":\"Du Monde\",\"company\":null,\"phone\":null,\"street1\":\"1 Tchoupitoulas St\",\"street2\":null,\"city\":\"New Orleans\",\"region\":\"LA\",\"postal_code\":\"70115\",\"country\":\"US\",\"nickname\":\"Home\",\"email\":null,\"vat_number\":null,\"id\":\"ljvmmbk9e1as\",\"account_id\":\"ljvmmbjchtgs\",\"created_at\":\"2019-09-19T22:45:59Z\",\"updated_at\":\"2019-09-19T22:45:59Z\"}],\"custom_fields\":[],\"hosted_login_token\":\"PSvcHow5H4HGEGKTfHXadLNoDcRaDVMK\",\"created_at\":\"2019-09-19T22:45:59Z\",\"updated_at\":\"2019-09-19T22:45:59Z\",\"deleted_at\":null}"
+
+Benchmark.bm do |benchmark|
```
bhelx: I committed this because I might expand this in the future.

lib/recurly/schema.rb
```diff
@@ -46,71 +45,108 @@ def self.get_recurly_class(type)
...
-        t = type == Array ? options[:item_type] : type
-        t.is_a?(Class) || t == :Boolean
+      def cast(value)
+        self.recurly_class.cast(value)
       end
     end
 
-    private_constant :List
-    private_constant :Attribute
+    class ArrayAttribute < Attribute
```
bhelx: Each Attribute type now has it's own class which contains the logic for validating and casting. This means we don't have to run through conditionals to find the right code to validate or cast a value. it's reduced to a hash lookup.

bhelx: These are also easier to test or change in isolation.

lib/recurly/schema/json_parser.rb
```diff
@@ -61,8 +66,11 @@ def self.recurly_class(type)
           if klazz.ancestors.include?(Resource)
             klazz
           else
-            # TODO might want to throw an error?
-            nil
```
bhelx: Picking up some TODOs in this commit

PR #512, opened 2019-09-20
Update travis options

bhelx:


.travis.yml
```diff
@@ -8,22 +8,25 @@ before_install:
...
+  - jruby-9.1
   - jruby-9000
   - 1.9.3-p551
   - 2.0.0-p648
-  - 2.1.10
-  - 2.2.9
-  - 2.3.6
-  - 2.4.3
-  - 2.5.0
+  - 2.1
```
bhelx: Not specifying the patch number here installs the latest patch.

.travis.yml
```diff
@@ -8,22 +8,25 @@ before_install:
...
   - XML=nokogiri
 matrix:
   allow_failures:
     - rvm: 1.9.3-p551
     - rvm: 2.0.0-p648
-    - rvm: 2.1.10
-    - rvm: 2.2.9
+    - rvm: 2.1
+    - rvm: 2.2
+    - rvm: 2.3
```
bhelx: Adding 2.3 to the allow_failures as it has been EOL'd : https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/

.travis.yml
joannasese: was travis breaking down because of outdated patches?

bhelx:
Good to go, just some small updates.

PR #537, opened 2019-11-27
Disable searching ancestors when looking up constants

### LINKED ISSUE #536: Undefined method cast
_Warning_: Github issues are not an official support channel for Recurly. If you have an urgent request we suggest you contact us through an official channel: support@recurly.com or https://recurly.zendesk.com

**Describe the bug**

We have an `ActiveRecord` class called `Invoice` and for some reason this was sometimes breaking specs. To fix this, I added:
```
elsif type == :Invoice
  Resources::Invoice
```
to `lib/recurly/schema.rb:46` and that got specs passing correctly.

**Expected behavior**
Gem should not be trying to cast the value of an external class.
**Your Environment**
- Which version of this library are you using? - 3.1

douglasmiller:
Resolves #536

The `const_defined?` and `const_get` methods will search the full ancestor tree and will provide a result for the top most matching constant when the `inherit` parameter is true. This causes issues when an application has constants defined on the top level scope that share names with Recurly constants.

The issue was verified to exist using the `bin/pry` script:
```ruby
[1] pry(main)> require 'recurly'
=> true
[2] pry(main)> Recurly::Schema.get_recurly_class(:Invoice)
=> Recurly::Resources::Invoice
[3] pry(main)> class Invoice; end
=> nil
[4] pry(main)> Recurly::Schema.get_recurly_class(:Invoice)
=> Invoice
```

After the change, the same test produces the correct result:
```ruby
[1] pry(main)> require 'recurly'
=> true
[2] pry(main)> Recurly::Schema.get_recurly_class(:Invoice)
=> Recurly::Resources::Invoice
[3] pry(main)> class Invoice; end
=> nil
[4] pry(main)> Recurly::Schema.get_recurly_class(:Invoice)
=> Recurly::Resources::Invoice
```

### DIFF
--- a/lib/recurly/schema.rb
+++ b/lib/recurly/schema.rb
@@ -42,11 +42,11 @@ module Recurly
       raise ArgumentError, "#{type.inspect} must be a symbol but is a #{type.class}" unless type.is_a?(Symbol)
 
       if type == :Address
-        Resources::Address
-      elsif Requests.const_defined?(type)
-        Requests.const_get(type)
-      elsif Recurly::Resources.const_defined?(type)
-        Resources.const_get(type)
+        Recurly::Resources::Address
+      elsif Recurly::Requests.const_defined?(type, false)
+        Recurly::Requests.const_get(type, false)
+      elsif Recurly::Resources.const_defined?(type, false)
+        Recurly::Resources.const_get(type, false)
       else
         raise ArgumentError, "Recurly type '#{type}' is unknown"
       end
--- a/lib/recurly/schema/json_parser.rb
+++ b/lib/recurly/schema/json_parser.rb
@@ -61,17 +61,10 @@ module Recurly
         Resources::Page
       else
         type_camelized = type.split("_").map(&:capitalize).join
-        if Resources.const_defined?(type_camelized)
-          klazz = Resources.const_get(type_camelized)
-          if klazz.ancestors.include?(Resource)
-            klazz
-          else
-            if Recurly::STRICT_MODE
-              raise ArgumentError, "Could not find Recurly Resource responsible for key #{type}"
-            else
-              nil
-            end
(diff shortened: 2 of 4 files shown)


lib/recurly/schema.rb
```diff
@@ -42,11 +42,11 @@ def self.get_recurly_class(type)
       raise ArgumentError, "#{type.inspect} must be a symbol but is a #{type.class}" unless type.is_a?(Symbol)
 
       if type == :Address
-        Resources::Address
-      elsif Requests.const_defined?(type)
-        Requests.const_get(type)
-      elsif Recurly::Resources.const_defined?(type)
-        Resources.const_get(type)
+        Recurly::Resources::Address
+      elsif Recurly::Requests.const_defined?(type, false)
```
bhelx: Do we need to do the same to this call? https://github.com/recurly/recurly-client-ruby/blob/65f65171cc2cbb6db99f2ea4c8b327ae12577fcc/lib/recurly/schema/json_parser.rb#L64

douglasmiller: Done

lib/recurly/schema/json_parser.rb
```diff
@@ -61,16 +61,11 @@ def self.recurly_class(type)
...
-            klazz
-          else
-            if Recurly::STRICT_MODE
-              raise ArgumentError, "Could not find Recurly Resource responsible for key #{type}"
-            else
-              nil
-            end
+        if Resources.const_defined?(type_camelized, false)
+          Resources.const_get(type_camelized, false)
+        else
```
bhelx: What do you think about making this `elsif Recurly::STRICT_MODE` then it would not have an else clause (returning nil)?

bhelx:
👍

PR #549, opened 2020-02-03
Loosen version restriction on faraday

### LINKED ISSUE #548: Update Faraday Version Maximum
Hello! Are there any plans underway to update the max-version constraint on the Faraday dependency? I see here: https://github.com/recurly/recurly-client-ruby/pull/518 that a max constraint of 0.16 was added to get specs to pass. Would love for this to be revisited soon so that the recurly gem does not prevent me from adding gems that use higher versions of faraday.

bhelx:
Resolves #548

Allows consumers of library to use newer versions of Faraday (but still restricting use of 1.0.0).


bhelx:
I tested this locally as well on some integration tests with `0.17.3`. Appears to work.

PR #552, opened 2020-02-10
Latest v2019-10-10 Changes

bhelx:
* More documentation
* `Account#has_*` flag attributes
* Include `RateLimitedError`
* Gateway Token Support


lib/recurly/client.rb
```diff
@@ -126,30 +126,30 @@ def handle_response!(request, faraday_resp)
       response = HTTP::Response.new(faraday_resp, request)
       raise_api_error!(response) unless (200...300).include?(response.status)
       resource = if response.body
```
bhelx: seems that rufo going back and forth about which syntax to use.

douglasmiller: Saw that and just verified locally that `0.7.0` formats this way, and `0.10.0` is the latest.

I bet if you bumped the version in the `gemspec` then all would be happy again

PR #553, opened 2020-02-11
Add convert_trial() to Subscription

joannasese:
To convert trial to paid subscription with 3ds token:
```ruby
subscription.convert_trial(token)
```

To convert trial for non-3ds where subscription has billing info:
```ruby
subscription.convert_trial
```

To convert trial to paid subscription when `transaction_type = moto`:
```ruby
subscription.convert_trial_moto
```

### DIFF
--- a/lib/recurly/api.rb
+++ b/lib/recurly/api.rb
@@ -18,7 +18,7 @@ module Recurly
     @@base_uri = "https://api.recurly.com/v2/"
     @@valid_domains = [".recurly.com"]
 
-    RECURLY_API_VERSION = '2.24'
+    RECURLY_API_VERSION = '2.25'
 
     FORMATS = Helper.hash_with_indifferent_read_access(
       'pdf' => 'application/pdf',
--- a/lib/recurly/subscription.rb
+++ b/lib/recurly/subscription.rb
@@ -174,6 +174,35 @@ module Recurly
       sub.tap {|e| e.currency = currency} if sub.is_a? Subscription
     end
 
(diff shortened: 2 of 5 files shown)


bhelx:
I think we want something like this:

```ruby
    # Convert free trial to paid subscription when transaction_type is "moto"
    # which stands for "Mail Order Telephone Order".
    # 
    # @return true
    def convert_trial_moto
      builder = XML.new("<subscription/>")
      builder.add_element('transaction_type', "moto")
      reload API.put("#{uri}/convert_trial", builder.to_s)
      true
    end

    # Convert free trial to paid subscription. Optionally use a 3ds token.
    # 
    # @param billing_info [String] three_d_secure_action_result_token_id 
    #   returned by Recurly.js referencing the result of the 3DS authentication for PSD2.
    # @return true when payment is accepted
    def convert_trial(three_d_secure_action_result_token_id = nil)
      body = if three_d_secure_action_result_token_id
        builder = Recurly::XML.new("<subscription/>")
        account = builder.add_element('account')
        billing_info = account.add_element('billing_info')
        billing_info.add_element('three_d_secure_action_result_token_id', three_d_secure_action_result_token_id)
        builder.to_s
      end

      reload API.put("#{uri}/convert_trial", body)
      true
    end
```

So now you have 3 ways to call it:

```ruby

# For non-3ds where billing info is already there to be converted
subscription.convert_trial

# Optionally pass in a token if we need 3ds auth
subscription.convert_trial(3ds_token)

# If we want to explicitly skip the authorization step (like for a MOTO order)
subscription.convert_trial_moto
```

PR #560, opened 2020-03-09
Adding #first and #count methods to Pager

douglasmiller:
Adds the `first` and `count` methods to the `Pager` class.

The `first` method will perform a request to the API with the `limit` set to 1 and will only return the first result from the server.

The `count` method will perform a `HEAD` request to the API for the appropriate endpoint. The value of the response header, `recurly-total-records`, will be returned. This is the total number of records that will be returned by the pager.

### DIFF
--- a/lib/recurly/http.rb
+++ b/lib/recurly/http.rb
@@ -4,7 +4,7 @@ module Recurly
       attr_accessor :status, :body, :request,
         :request_id, :rate_limit, :rate_limit_remaining,
         :rate_limit_reset, :date, :proxy_metadata,
-        :content_type
+        :content_type, :total_records
 
       def initialize(resp, request)
         @request = Request.new(request.method, request.path, request.body)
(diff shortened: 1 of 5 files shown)


lib/recurly/http.rb
```diff
@@ -13,6 +13,7 @@ def initialize(resp, request)
         @rate_limit = resp["x-ratelimit-limit"].to_i
         @rate_limit_remaining = resp["x-ratelimit-remaining"].to_i
         @rate_limit_reset = Time.at(resp["x-ratelimit-reset"].to_i).to_datetime
+        @total_records = resp["recurly-total-records"].to_i
```
bhelx, line 16: We may want to put this in a conditional or else it will come out as `0`. perhaps:
```ruby
@total_records = resp["recurly-total-records"]&.to_i
```

douglasmiller: Updated

PR #568, opened 2020-03-22
Replace Faraday gem with Net::HTTP, add connection pooling & keep-alive, update CA roots

isaachall:
Removes the Faraday gem dependency and uses ruby's built-in `Net::HTTP`. This allows the Recurly client to have more predictable control over the API requests, enable keep alive consistently, and retry requests for some common network exceptions. Keep alive will skip the SSL handshake on additional requests, which will greatly speed up the performance of the API. The automatic retries will allow the API to be more reliable. Please note: this change will only retry requests whose bodies were not submitted to the server -- these requests all failed before the body was successfully sent.

The second commit updates the bundled root CA certificate with Mozilla's root CA bundle. This will ensure that we can properly validate our root public certificate with a range of trusted certificate authorities.

This resolves all classes of problems similar to #548


lib/recurly/client.rb
```diff
@@ -118,18 +119,59 @@ def delete(path, **options)
...
+      request["Authorization"] = "Basic #{Base64.encode64(@api_key)}".chomp
+      request["User-Agent"] = "Recurly/#{VERSION}; #{RUBY_DESCRIPTION}"
+
+      # TODO this is undocumented until we finalize it
+      options[:headers].each { |header, v| request[header] = v } if options[:headers]
+
+      connection_pool.with_connection do |http|
+        http.open_timeout = options[:open_timeout] || 20
+        http.read_timeout = options[:read_timeout] || 60
+
```
bhelx: We need to plug in the logger somewhere for debugging purposes. If we drop it here, that will work. I used this code:

```
http.set_debug_output(logger)
```

bhelx:
```
/app/clients/ruby/lib/recurly/client.rb:137: warning: Net::HTTP#set_debug_output called after HTTP started
```

Perhaps it can only be set only once per started instance.

lib/recurly/client.rb
```diff
@@ -42,17 +46,18 @@ class Client
...
 
       # execute block with this client if given
       yield(self) if block_given?
     end
 
     def next_page(pager)
-      req = HTTP::Request.new(:get, pager.next, nil)
-      faraday_resp = run_request(req, headers)
-      handle_response! req, faraday_resp
+      path = URI(pager.next).path
```
bhelx: Is there a reason we need to parse into URI first? This line drops the query params:

```ruby
irb(main):003:0> URI("/accounts?cursor=xyz&limit=20&sort=created_at").path
=> "/accounts"
```

I think it might work to pass the `next` path directly into the Get constructor:

```ruby
request = Net::HTTP::Get.new path
```

Or if you need to parse the URI, there might be a way to reconstruct the request.

lib/recurly/client.rb
bhelx: Also, we need to gate this only for certain logger levels.

lib/recurly/client.rb
```diff
@@ -68,44 +73,40 @@ def pager(path, **options)
...
       path = scope_by_site(path, **options)
-      request = HTTP::Request.new(:put, path)
+      request = Net::HTTP::Put.new path
       if request_data
         request_class.new(request_data).validate!
-        logger.info("PUT BODY #{JSON.dump(request_data)}")
-        request.body = JSON.dump(request_data)
+        json_body = JSON.dump(request_data)
+        logger.info("PUT BODY #{json_body}")
+        request.set_content_type(JSON_CONTENT_TYPE)
```
bhelx: It seems the server wants the content-type to be json regardless of whether or not there is a body. Moving this line to line `93` fixes.

lib/recurly/client.rb
isaachall: I added back the logger and added specs to make sure it gets called with the correct levels.

lib/recurly/client.rb
isaachall: Updated in the latest commit, thanks!

lib/recurly/client.rb
isaachall: Makes sense, updated.

lib/recurly/client.rb
```diff
@@ -118,18 +124,67 @@ def delete(path, **options)
...
+          if ex.kind_of?(Net::OpenTimeout) || ex.kind_of?(Errno::ETIMEDOUT)
+            raise Recurly::Errors::TimeoutError, "Request timed out"
+          end
+
+          raise Recurly::Errors::ConnectionFailedError, "Failed to connect to Recurly: #{ex.message}"
+        rescue Net::ReadTimeout, Timeout::Error
+          raise Recurly::Errors::TimeoutError, "Request timed out"
+        rescue OpenSSL::SSL::SSLError => ex
+          raise Recurly::Errors::SSLError, ex.message
+        end
```
bhelx: Should we have a default rescue here? Also should we catch and translate SocketError as well?
```
Traceback (most recent call last):
	13: from scripts/v2019-10-10/ruby/create_account.rb:34:in `<main>'
	12: from /app/clients/ruby/lib/recurly/client/operations.rb:147:in `create_account'
	11: from /app/clients/ruby/lib/recurly/client.rb:90:in `post'
	10: from /app/clients/ruby/lib/recurly/client.rb:135:in `run_request'
	 9: from /app/clients/ruby/lib/recurly/connection_pool.rb:19:in `with_connection'
	 8: from /app/clients/ruby/lib/recurly/client.rb:141:in `block in run_request'
	 7: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:925:in `start'
	 6: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:930:in `do_start'
	 5: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:945:in `connect'
	 4: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/timeout.rb:103:in `timeout'
	 3: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/timeout.rb:93:in `block in timeout'
	 2: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:947:in `block in connect'
	 1: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:947:in `open'
/usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:947:in `initialize': getaddrinfo: Name or service not known (SocketError)
	12: from scripts/v2019-10-10/ruby/create_account.rb:34:in `<main>'
	11: from /app/clients/ruby/lib/recurly/client/operations.rb:147:in `create_account'
	10: from /app/clients/ruby/lib/recurly/client.rb:90:in `post'
	 9: from /app/clients/ruby/lib/recurly/client.rb:135:in `run_request'
	 8: from /app/clients/ruby/lib/recurly/connection_pool.rb:19:in `with_connection'
	 7: from /app/clients/ruby/lib/recurly/client.rb:141:in `block in run_request'
	 6: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:925:in `start'
	 5: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:930:in `do_start'
	 4: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:945:in `connect'
	 3: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/timeout.rb:103:in `timeout'
	 2: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/timeout.rb:93:in `block in timeout'
	 1: from /usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:946:in `block in connect'
/usr/local/rbenv/versions/2.6.3/lib/ruby/2.6.0/net/http.rb:949:in `rescue in block in connect': Failed to open TCP connection to v3.recurly.con:443 (getaddrinfo: Name or service not known) (SocketError)
```

lib/recurly/client.rb
```diff
@@ -139,25 +194,14 @@ def handle_response!(request, faraday_resp)
...
     def raise_api_error!(response)
-      error = JSONParser.parse(self, response.body)
-      error_class = Errors::APIError.error_class(error.type)
-      raise error_class.new(response, error)
+      if response.content_type.include?(JSON_CONTENT_TYPE)
+        error = JSONParser.parse(self, response.body)
+        error_class = Errors::APIError.error_class(error.type)
+        raise error_class.new(response, error)
+      else
+        raise Recurly::Errors::InvalidResponseError, "Unexpected content type: #{http_response.content_type}"
```
bhelx: I believe this line needs to be changed to as `http_response` doesn't exist in this context:

```ruby
raise Recurly::Errors::InvalidResponseError, "Unexpected content type: #{response.content_type}"
```

lib/recurly/client.rb
isaachall: 👍

lib/recurly/client.rb
isaachall: Thanks, my mistake as I refactored to reduce complexity. I added some specs around these conditions to better flush this out.

PR #579, opened 2020-04-07
Set an Idempotency-Key header, retry GET requests after 5xx errors

isaachall:
After 2020-04-09, API v3 will support an `Idempotency-Key` header so we can safely resubmit POST, PUT, and DELETE requests. Adding this header makes the requests more resilient.

Adds one retry for GET requests that return a 5xx server error. These requests are somewhat likely to succeed on a retry.

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -2,6 +2,7 @@ require "logger"
 require "erb"
 require "net/https"
 require "base64"
+require "securerandom"
 require_relative "./schema/json_parser"
 require_relative "./schema/file_parser"
 
@@ -17,6 +18,8 @@ module Recurly
     ]
     JSON_CONTENT_TYPE = "application/json"
     MAX_RETRIES = 3
+
+    BASE36_ALPHABET = ("0".."9").to_a + ("a".."z").to_a
 
     # Initialize a client. It requires an API key.
     #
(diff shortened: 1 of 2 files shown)


lib/recurly/client.rb
```diff
@@ -76,8 +79,16 @@ def get(path, **options)
       path = scope_by_site(path, **options)
       request = Net::HTTP::Get.new path
       set_headers(request, options[:headers])
-      http_response = run_request(request, options)
-      handle_response! request, http_response
+
+      retries = 0
```
bhelx, line 83: What do you think about moving this logic to `run_request` or find a way to share it? `next_page` needs it for instance as that's what the Pager uses. It'll make it easier to turn on for other methods in the future as well.

lib/recurly/client.rb
```diff
@@ -76,8 +79,16 @@ def get(path, **options)
...
+      begin
+        http_response = run_request(request, options)
+        handle_response! request, http_response
+      rescue Recurly::Errors::UnavailableError, Recurly::Errors::InternalServerError
+        retries += 1
+        retry if retries < 2 # Retry 5xx once for GET requests only
```
bhelx, line 89: Should we add sleeping before retrying?

lib/recurly/client.rb
```diff
@@ -166,11 +177,21 @@ def set_headers(request, additional_headers = {})
       request["Accept"] = "application/vnd.recurly.#{api_version}".chomp # got this method from operations.rb
       request["Authorization"] = "Basic #{Base64.encode64(@api_key)}".chomp
       request["User-Agent"] = "Recurly/#{VERSION}; #{RUBY_DESCRIPTION}"
+      request["Idempotency-Key"] ||= generate_idempotency_key unless request.is_a?(Net::HTTP::Get)
```
bhelx, line 180: Do we need the keys for DELETEs and HEADs?

lib/recurly/client.rb
isaachall: I think we should add some sleeps for certain network conditions, although it will pretty much be a guess :) I'm less concerned about sleeping in between server 5xx responses since the request is likely to route to a different backend process that may not have the same issue.

lib/recurly/client.rb
isaachall: Valid! I found a way to move it to `#run_request` that doesn't add much complexity.

lib/recurly/client.rb
isaachall: I changed it so it does not send on GETs and HEADs. DELETEs are a funny one; we have some endpoints that trigger refunds via a delete request. So, we're allowing the idempotency header on those requests.

speza:
@isaachall has this new header not been documented against the V3 API spec? I'm struggling to find any references to it other than in this Git issue.

PR #580, opened 2020-04-09
Use github pages for docs

bhelx:
Rubydoc.info consistently loses our project. It's very disruptive and we have little control over the process. This is an attempt to host the rubydocs with github pages.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,28 @@
+on:
+  release:
+    types:
+      - published
+name: Documentation
+jobs:
+  build:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: '2.6'
+
+    - name: Build Library and Docs
+      run: ./scripts/build
+
+    - name: Deploy
+      if: success()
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.6'
+
+    - name: Build Library and Docs
+      run: ./scripts/build
```
bhelx, line 18: This calls bundle install and the yard doc generator. Do I need the bundle install part?

chrissrogers: If doc generation requires the bundle, then yep

README.md
```diff
@@ -7,7 +7,7 @@ This repository houses the official ruby client for Recurly's V3 API.
 
 ## Reference Documentation
 
-Getting Started Guide and reference documentation can be found on [rubydoc.info](https://www.rubydoc.info/github/recurly/recurly-client-ruby/).
+Getting Started Guide and reference documentation can be found on [on Github Pages](https://recurly.github.io/recurly-client-ruby/).
```
chrissrogers, line 10: oops, "on on"

PR #581, opened 2020-04-13
Fixing the omission of query params

douglasmiller:
Only the `Pager` requests were including the optional query parameters when performing API requests.

This update moves the logic to handle query params from the `Pager` to the `Client` and consolidates some resulting portions of the `Pager` and `Client`

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -53,14 +53,6 @@ module Recurly
       yield(self) if block_given?
     end
 
-    def get_resource_count(path, options)
-      request = Net::HTTP::Head.new build_path(path, options)
-      set_headers(request)
-      http_response = run_request(request)
-      resource = handle_response!(request, http_response)
-      resource.get_response.total_records
-    end
-
     protected
 
     def pager(path, **options)
@@ -69,6 +61,14 @@ module Recurly
         path: path,
         options: options,
       )
+    end
+
+    def head(path, **options)
+      path = scope_by_site(path, **options)
+      request = Net::HTTP::Head.new build_path(path, options)
+      set_headers(request, options[:headers])
+      http_response = run_request(request, options)
+      handle_response! request, http_response
     end
 
     def get(path, **options)
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -53,14 +53,6 @@ module Recurly
       yield(self) if block_given?
     end
 
-    def next_page(url)
-      path = extract_path(url)
-      request = Net::HTTP::Get.new path
-      set_headers(request)
-      http_response = run_request(request)
-      handle_response! request, http_response
-    end
-
     def get_resource_count(url)
       request = Net::HTTP::Head.new url
       set_headers(request)
@@ -72,7 +64,6 @@ module Recurly
     protected
 
     def pager(path, **options)
-      path = scope_by_site(path, **options)
       Pager.new(
         client: self,
         path: path,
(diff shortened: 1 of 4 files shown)


lib/recurly/client.rb
```diff
@@ -260,19 +259,13 @@ def set_api_key(api_key)
     end
 
     def scope_by_site(path, **options)
-      if site = site_id || options[:site_id]
+      if (site = site_id || options[:site_id]) && !path.start_with?("/sites/#{site}")
```
bhelx, line 262: Perhaps let's add a comment about what this is doing but that it's not officially supported.

spec/recurly/pager_spec.rb
```diff
@@ -73,15 +67,9 @@
 
     describe "#each_page" do
       it "should present an enumerator pages" do
-        expect(client).to receive(:next_page).exactly(3).times
-        allow(client).to receive(:next_page).and_return(more_response)
-        page_num = 0
+        expect(client).to receive(:get).exactly(3).times
+        allow(client).to receive(:get).and_return(more_response, more_response, done_response)
```
douglasmiller, line 71: Cleaned up some unnecessary verbosity here

spec/recurly/pager_spec.rb
```diff
@@ -92,21 +80,24 @@
     end
     describe "#each" do
       it "should present an enumerator pages" do
-        expect(client).to receive(:next_page).exactly(3).times
-        allow(client).to receive(:next_page).and_return(more_response)
-        item_num = 0
+        expect(client).to receive(:get).exactly(3).times
+        allow(client).to receive(:get).and_return(more_response, more_response, done_response)
```
douglasmiller, line 84: Cleaned up some unnecessary verbosity here

lib/recurly/client.rb
```diff
@@ -53,16 +53,8 @@ def initialize(api_key:, site_id: nil, subdomain: nil, **options)
...
-      path = extract_path(url)
-      request = Net::HTTP::Get.new path
-      set_headers(request)
-      http_response = run_request(request)
-      handle_response! request, http_response
-    end
-
-    def get_resource_count(url)
-      request = Net::HTTP::Head.new url
+    def get_resource_count(path, options)
```
bhelx, line 56: Can't we also get rid of this? and move it into the `Pager`? It could call a `head` method in the client.

lib/recurly/client.rb
```diff
@@ -259,18 +249,33 @@ def set_api_key(api_key)
       @api_key = api_key
     end
 
-    def scope_by_site(path, **options)
-      if site = site_id || options[:site_id]
-        "/sites/#{site}#{path}"
+    def validate_options!(options)
+      options.keys.each do |key|
```
bhelx, line 253: What do you think about getting all bad keys at once, like this:

```ruby
invalid_options = options.keys - ALLOWED_OPTIONS
if invalid_options.any?
  # .. raise the error here
end
```

bhelx: yes

douglasmiller:
I rolled back the `params` enforcement as discussed elsewhere

lib/recurly/client.rb
```diff
@@ -53,45 +53,36 @@ def initialize(api_key:, site_id: nil, subdomain: nil, **options)
       yield(self) if block_given?
     end
 
-    def next_page(url)
```
bhelx, line 56: For anyone watching, we decided to refactor this functionality by making it private / shared.

bhelx: The Pager now calls the private API of Client

lib/recurly/pager.rb
```diff
@@ -7,24 +7,25 @@ def initialize(client:, path:, options: {})
...
       # Restore the @next url to the original
       @next = original_next
       @data.first
     end
 
     # Makes a HEAD request to the API to determine how many total records exist.
     def count
-      @client.get_resource_count(self.next)
+      resource = @client.send(:head, self.next, @options)
+      return resource.get_response.total_records
```
bhelx, line 28: nit: let's use implicit return here

PR #588, opened 2020-04-29
Implement Connection Pooling

bhelx:
Backporting the connection pooling code from V3 to ensure that connections are being properly reused. We should ensure that there have been no behavioral changes.

I've adjusted this so you must opt-in to it:

```
Recurly.use_connection_pool(true)
```

### DIFF
--- a/lib/recurly/api/net_http_adapter.rb
+++ b/lib/recurly/api/net_http_adapter.rb
@@ -5,6 +5,11 @@ module Recurly
   class API
     module Net
       module HTTPAdapter
+
+        def connection_pool
+          @connection_pool ||= Recurly::ConnectionPool.new
+        end
+
         # A hash of Net::HTTP settings configured before the request.
         #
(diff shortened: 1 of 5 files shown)


lib/recurly/api/net_http_adapter.rb
```diff
@@ -5,6 +5,10 @@ module Recurly
   class API
     module Net
       module HTTPAdapter
+        def connection_pool
+          @connection_pool ||= Recurly::ConnectionPool.new
```
bhelx, line 9: This maybe needs to be a class var

bhelx:
I'm holding off on this change because i'm not yet prepared to support in master. The branch will stay up and it should be safe to use for scripts. I would be cautious about using it in a long running program (such as a rails app or a server).

bhelx:
*Note*: Don't delete this branch as support is using this internally.

PR #590, opened 2020-05-28
Support the programmer passing their own logger

bhelx:
This adds a `logger` option when creating the Client. It also slims down some of the logging messages. By default we will set the logging level to `INFO` which only logs high level HTTP details:

```
I, [2020-05-28T18:43:13.164922 #119]  INFO -- Recurly: [Request] method="POST" path="/accounts"
I, [2020-05-28T18:43:13.362743 #119]  INFO -- Recurly: [Response] time_ms=197 status="201" method="POST" path="/accounts"
I, [2020-05-28T18:43:13.364346 #119]  INFO -- Recurly: [Request] method="DELETE" path="/accounts/code-0d18c3c49f7"
I, [2020-05-28T18:43:13.525843 #119]  INFO -- Recurly: [Response] time_ms=161 status="200" method="DELETE" path="/accounts/code-0d18c3c49f7"
```

Setting level `DEBUG` will log things like headers and request / response bodies.

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -148,12 +148,12 @@ module Recurly
           http.start unless http.started?
           log_attrs = {
             method: request.method,
-            path: request.path
+            path: request.path,
           }
           if @logger.level < Logger::INFO
             log_attrs[:request_body] = request.body
             # No need to log the authorization header
-            headers = request.to_hash.reject { |k,_| k&.downcase == 'authorization' }
+            headers = request.to_hash.reject { |k, _| k&.downcase == "authorization" }
             log_attrs[:request_headers] = headers
           end
 
@@ -333,8 +333,8 @@ module Recurly
     %i(debug info warn error fatal).each do |level|
       define_method "log_#{level}" do |tag, **attrs|
         @logger.send(level, "Recurly") do
-           msg = attrs.each_pair.map { |k,v| "#{k}=#{v.inspect}" }.join(" ")
-           "[#{tag}] #{msg}"
+          msg = attrs.each_pair.map { |k, v| "#{k}=#{v.inspect}" }.join(" ")
+          "[#{tag}] #{msg}"
         end
       end
     end
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -46,7 +46,7 @@ module Recurly
     #
     # @param api_key [String] The private API key
     # @param logger [Logger] A logger to use. Defaults to creating a new STDOUT logger with level WARN.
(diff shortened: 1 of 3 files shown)


lib/recurly/client.rb
```diff
@@ -45,12 +45,28 @@ class Client
...
+
+      if logger.nil?
+        @logger = Logger.new(STDOUT).tap do |l|
+          l.level = Logger::INFO
+        end
+      else
+        @logger = logger
+      end
+
+      if @logger.level < Logger::INFO
```
bhelx, line 61: I'm considering requiring an environment variable being set to do this. There are security implications to someone accidentally passing in a debug logger.

lib/recurly/client.rb
```diff
@@ -45,12 +45,28 @@ class Client
...
       set_site_id(site_id, subdomain)
       set_api_key(api_key)
-      set_options(options)
+
+      if logger.nil?
+        @logger = Logger.new(STDOUT).tap do |l|
+          l.level = Logger::INFO
+        end
+      else
+        @logger = logger
```
bhelx, line 58: I should probably do some basic validation of this logger. And make the `log_<level>` calls more resilient to failures.

lib/recurly/client.rb
```diff
@@ -45,12 +45,28 @@ class Client
...
-    def initialize(api_key:, site_id: nil, subdomain: nil, **options)
+    # @param logger [Logger] A logger to use. Defaults to creating a new STDOUT logger with level INFO.
+    def initialize(api_key:, site_id: nil, subdomain: nil, logger: nil, **options)
       set_site_id(site_id, subdomain)
       set_api_key(api_key)
-      set_options(options)
+
+      if logger.nil?
+        @logger = Logger.new(STDOUT).tap do |l|
+          l.level = Logger::INFO
```
bhelx, line 55: Wondering if maybe we should default to `WARN` instead of info?

lib/recurly/client.rb
```diff
@@ -45,12 +45,31 @@ class Client
...
-    def initialize(api_key:, site_id: nil, subdomain: nil, **options)
+    # @param logger [Logger] A logger to use. Defaults to creating a new STDOUT logger with level WARN.
+    def initialize(api_key:, site_id: nil, subdomain: nil, logger: nil)
       set_site_id(site_id, subdomain)
       set_api_key(api_key)
-      set_options(options)
+
+      if logger.nil?
+        @logger = Logger.new(STDOUT).tap do |l|
+          l.level = Logger::WARN
```
douglasmiller, line 55: 👍 I was going to comment that `INFO` might not be a preferred default as it could result in a non-trivial increase to log output.

PR #597, opened 2020-06-22
Allow :headers to be included in operations

douglasmiller:
Fixes bug that prevents `:headers` from being included in an operation's `options`

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -20,6 +20,7 @@ module Recurly
     MAX_RETRIES = 3
     LOG_LEVELS = %i(debug info warn error fatal).freeze
     BASE36_ALPHABET = (("0".."9").to_a + ("a".."z").to_a).freeze
+    REQUEST_OPTIONS = [:headers].freeze
 
     # Initialize a client. It requires an API key.
     #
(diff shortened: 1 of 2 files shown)


spec/recurly/client_spec.rb
```diff
@@ -54,6 +54,15 @@
         expect(net_http).to receive(:request).and_return(response)
         _account = subject.get_account(account_id: "code-benjamin-du-monde")
       end
+
+      it "should include custom headers in each request" do
+        expected = hash_including("Custom-Header" => "custom-value", "Accept" => /application\/vnd\.recurly/, "Content-Type" => "application/json", "User-Agent" => /Recurly\//)
```
bhelx, line 59: Need to assert this

douglasmiller: Yeah, I based this spec off of the one above it and neither is actually asserting the `expected`. I'll update it

lib/recurly/client.rb
```diff
@@ -204,16 +205,16 @@ def run_request(request, options = {})
     end
 
     def set_headers(request, additional_headers = {})
+      # TODO this is undocumented until we finalize it
```
douglasmiller, line 208: Moved this to the top to prevent overwriting the standard headers

PR #600, opened 2020-06-26
Update error mapping

douglasmiller:
Forward planning for v4 ruby client.


lib/recurly/errors/network_errors.rb
```diff
@@ -6,5 +6,17 @@ class TimeoutError < NetworkError; end
     class ConnectionFailedError < NetworkError; end
     class SSLError < NetworkError; end
     class UnavailableError < NetworkError; end
+
+    # Missing generic 5XX -> UnavailableError
```
douglasmiller, line 10: Need to find a way to work around this

douglasmiller:
Not necessary

PR #613, opened 2020-07-15
Add support for Avalara for Communications

bhelx:
Only used by customers who have "Avalara for Communications" integrations.


spec/recurly/add_on_spec.rb
```diff
@@ -34,6 +34,17 @@
       add_on.add_on_code.must_equal "marketing_email"
     end
 
+    it "must return a avalara info when available" do
```
joannasese, line 37: Super nit: `it must return avalara info when available` or `it must return an avalara info when available`

PR #616, opened 2020-07-22
Updating error mapping based on status code

douglasmiller:
This is an update to the `4.x` version of the client.

Updates the client to classify non-standard error responses from Recurly API via the HTTP status code mapping provided in `Recurly::Errors::ERROR_MAP`.


lib/recurly/errors.rb
```diff
@@ -1,6 +1,6 @@
 module Recurly
   module Errors
-    class APIError < StandardError
+    class ApiError < StandardError
```
bhelx, line 3: We should leave this the same if possible

PR #634, opened 2020-09-10
New endpoint to verify an account's billing information

joannasese:
If the account has billing information for anything other than a credit card, this will return 422, with an appropriate error message. If successful, this will return 200, with the transaction's data.
`gateway_code` (optional) is the code for the gateway to use for the verification. If unspecified, a gateway will be selected using the normal rules.

Example:
```ruby
# verify without specifying gateway code to use 
billing_info.verify()
# verify with a specific gateway code
billing_info.verify({gateway_code: "gateway-code"})
```


joannasese:
when you change your base branch, do it correctly

PR #644, opened 2020-10-28
Support item-specific coupons

joannasese:
Add `item_codes` and `applies_to_all_items` to Coupon

Example
```ruby
# Create coupon with `item_codes`
coupon = Recurly::Coupon.create!(
  coupon_code:      'special',
  name:             'Special',
  discount_type:    'dollars',
  discount_in_cents: 2_00,
  
  # `applies_to_all_items` defaults to false, so next line is optional
  applies_to_all_items: false,
  item_codes: ['one_item', 'two_item'],
)
```

### DIFF
--- a/lib/recurly/coupon.rb
+++ b/lib/recurly/coupon.rb
@@ -46,6 +46,8 @@ module Recurly
       unique_template_code
       free_trial_amount
       free_trial_unit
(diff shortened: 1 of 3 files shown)


douglasmiller (approved):
👍

PR #648, opened 2020-11-10
Expose last_two on BillingInfo class for IBAN accounts and repair tests

joannasese:
Response for https://developers.recurly.com/api-v2/v2.29/#operation/createAccountsBillingInfoBankAccount endpoint using `iban` should include `last_two`, similar to `last_four` when `account_number` is used. This patch exposes the `last_two` field. 

Test has also been fixed to include `last_two` and remove `iban`, which was included erroneously.

### DIFF
--- a/spec/recurly/billing_info_spec.rb
+++ b/spec/recurly/billing_info_spec.rb
@@ -62,7 +62,7 @@ XML
       )
       billing_info = BillingInfo.find 'sepa1234567890'
       billing_info.name_on_account.must_equal 'Account Name'
-      billing_info.iban.must_equal 'US1234567890'
+      billing_info.last_two.must_equal '06'
     end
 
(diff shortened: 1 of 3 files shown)


spec/recurly/billing_info_spec.rb
```diff
@@ -62,7 +62,7 @@
       )
       billing_info = BillingInfo.find 'sepa1234567890'
       billing_info.name_on_account.must_equal 'Account Name'
-      billing_info.iban.must_equal 'US1234567890'
```
jguidry-recurly, line 65: Given the removal of `iban`, it seems like test name should be updated. Maybe "must return an account's billing info as last two when available"?

joannasese: Good call out. I made some changes.

jguidry-recurly (approved):

PR #650, opened 2020-11-13
Expose item_state and external_sku in AddOn class

joannasese:
Because `item_state` and `external_sku` are included in the xml response for item-based add-ons, it would be good practice to expose this fields to the programmer.


jguidry-recurly (approved):
LGTM

PR #651, opened 2020-11-17
Latest Changes for 2019-10-10

douglasmiller:
- Adds the `state` parameter to the `list_account_coupon_redemptions` operation.
- Adds the `charge` parameter to the `terminate_subscription` operation so that a merchant can choose to not bill a customer for un-billed usage when terminating a subscription


joannasese (approved):
👍

Issue #652, opened 2020-11-18
Refund API V2 Error Details


marcmoy:
Hi, is there a way to get additional details about a `#<Recurly::API::UnprocessableEntity: 422 Unprocessable Entity (POST https://api.recurly.com/v2/invoices/..../refund)` error? The `Recurly::API::UnprocessableEntity` class does not provide a method to inspect the exact reason behind the error. Based on what I'm seeing in the Recurly dashboard, I'm noticing the refund could be rejected due to an existing chargeback, bank decline, or card stolen error. We would like to perform different actions in each case.

(Using release version `2.17.11`)

douglasmiller:
Hey @marcmoy, 

The `UnprocessableEntity` error extends `ClientError` which extends `ResponseError`. You should be able to obtain adequate information about the cause of the error from [the methods on `ResponseError`](https://github.com/recurly/recurly-client-ruby/blob/v2/lib/recurly/api/errors.rb#L6). Have you checked `UnprocessableEntity#symbol`, `UnprocessableEntity#description`, and/or `UnprocessableEntity#details`?

marcmoy:
@douglasmiller Yes I tried inspecting the error but not seeing the results I'm expecting:
```
irb(main):032:0> e
=> #<Recurly::API::UnprocessableEntity: 422 Unprocessable Entity (POST https://api.recurly.com/v2/invoices/114677553/refund)>
irb(main):033:0> e.symbol
=> nil
irb(main):034:0> e.description
=> nil
irb(main):035:0> e.details
=> nil
irb(main):036:0> e.message
=> "422 Unprocessable Entity (POST https://api.recurly.com/v2/invoices/114677553/refund)"
irb(main):037:0> e.methods
[...]

PR #653, opened 2020-11-23
Handle UnprocessableEntity error for Invoice#refund

joannasese:
Addresses https://github.com/recurly/recurly-client-ruby/issues/652

Provides more error details when a `Recurly::API::UnprocessableEntity` error is encountered. This patch exposes `Recurly::Transaction::Error` when a refund is not successful.


douglasmiller (approved):
👍

PR #659, opened 2020-12-21
Streamline release process:

joannasese:
- Create build-gem script that 1) checks out and pulls v2 branch and 2) builds ruby gem
- Create publish script, which will publish ruby gem - I wanted to keep building and publishing separate, but can combine them into one script if that's preferred.
- Add publish script file to bump2version configuration


douglasmiller:
Superseded by https://github.com/recurly/recurly-client-ruby/pull/690

PR #667, opened 2021-02-22
Add tests for subscription add-ons

joannasese:
Adding test coverage for updating subscription add-ons.

### DIFF
--- a/spec/recurly/add_on_spec.rb
+++ b/spec/recurly/add_on_spec.rb
@@ -64,14 +64,5 @@ describe AddOn do
       add_on.must_be_instance_of AddOn
       add_on.tier_type.must_equal "tiered"
     end
-
-    it "must return an addon with tiered-pricing when available" do
-      plan = Plan.find 'plantfile'
-      add_ons = plan.add_ons
-
-      add_on = add_ons.first
-      add_on.must_be_instance_of AddOn
-      add_on.tier_type.must_equal "tiered"
-    end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/recurly/subscription_spec.rb
```diff
@@ -162,6 +162,36 @@
...
+
+      it "will update subscription add-on tiers" do
+        subscription = Subscription.find '5895ce632d88b801b7831b41baba4c8b'
+        subscription.subscription_add_ons.length.must_equal 2
+        subscription.subscription_add_ons[1].tiers.length.must_equal 2
+
+        subscription.update_attributes!(timeframe: "now")
+        subscription.subscription_add_ons[1].tiers.length.must_equal 3
+        subscription.subscription_add_ons[1].tiers[0].unit_amount_in_cents["USD"].must_equal 5500
+        subscription.subscription_add_ons[1].tiers[1].unit_amount_in_cents["USD"].must_equal 4999
```
joannasese, line 192: Would it be better practice to also place a check on the third tier?

spec/recurly/add_on_spec.rb
```diff
@@ -64,14 +64,5 @@
       add_on.must_be_instance_of AddOn
       add_on.tier_type.must_equal "tiered"
     end
-
```
joannasese, line 67: Removing what appears to be a duplicate

spec/recurly/subscription_spec.rb
efeygelson: Yea I think we may as well add a check for the third tier too

efeygelson (approved):

PR #670, opened 2021-03-04
Update operations.rb

phaedryx:
Fix typo in comments


douglasmiller:
Thank you for bringing this to our attention, @phaedryx. The `operations.rb` is a generated file, so we will have to update the source of this misspelling and re-generate the files.

PR #672, opened 2021-03-15
Fixes TypeError caused by attempts to Base64 encode nil @api_key values

alexfulsome:
When a Client is initialized with a nil `api_key` value (a Dotenv loading issue in my case), requests fail with an opaque TypeError when `#set_headers` tries to base64 encode it:

```shell
[2] Pry(main)> subscriptions.first
TypeError: no implicit conversion of nil into String
from /usr/local/var/rbenv/versions/2.5.7/lib/ruby/2.5.0/base64.rb:39:in `pack'
```
This explicitly casts `@api_key` to a string, which makes a blank value raise a `Recurly::Errors::UnauthorizedError` instead.

### DIFF
--- a/lib/recurly/client.rb
+++ b/lib/recurly/client.rb
@@ -331,7 +331,7 @@ module Recurly
     end
 
     def set_api_key(api_key)
-      @api_key = api_key
+      @api_key = api_key.to_s
     end
 
     def build_url(path, options)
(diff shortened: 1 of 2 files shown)


douglasmiller:
Good catch, @alexfulsome.

Let's try to save a trip to the API and add something like `raise ArgumentError, "'api_key' must be set to a non-nil value" if api_key.nil?` to the initializer prior to calling `set_api_key`

douglasmiller (approved):
👍

PR #690, opened 2021-04-08
Adding release scripts

douglasmiller:
Adding release scripts and automatic changelog generation.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Usage
+#
+# ./release
+#
+
+# publish
+# Clean up any leftover gems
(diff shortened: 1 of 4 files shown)


chrissrogers (approved):

scripts/release
```diff
@@ -0,0 +1,17 @@
...
+# ./release
+#
+
+# publish
+# Clean up any leftover gems
+rm -f ./*.gem
+# Build the new gem
+gem build recurly.gemspec
+# Push what should be the only gem present
+gem push ./*.gem
```
chrissrogers, line 16: Noting that we could specify `gem push ./recurly-${RELEASE_TAG}.gem`, however, the glob is just fine given the rm.

PR #699, opened 2021-04-30
Making #post allow a nil body

douglasmiller:
The protected `Client#post` previously required that `request_data` and `request_class` be provided. This update will make the `#post` method allow those parameters to be `nil` which matches the behavior of the `#put` method.


joannasese (approved):
Was in call with you when this was discussed. 👍

PR #700, opened 2021-05-06
Add scheduled subscription update webhook

jguidry-recurly:
https://developers.recurly.com/reference/webhooks/index.html#scheduled-subscription-update


lib/recurly/webhook.rb
```diff
@@ -37,6 +37,7 @@ module Webhook
     autoload :SubscriptionResumedNotification,      'recurly/webhook/subscription_resumed_notification'
     autoload :CanceledSubscriptionNotification,     'recurly/webhook/canceled_subscription_notification'
     autoload :ScheduledSubscriptionPauseNotification,'recurly/webhook/scheduled_subscription_pause_notification'
+    autoload :ScheduledSubscriptionUpdate,          'recurly/webhook/scheduled_subscription_update_notification'
```
joannasese: Needs to be `autoload :ScheduledSubscriptionUpdateNotification`

joannasese: Tests should pass after that's fixed

jguidry-recurly: Would you believe that I actually have this fixed in an open buffer, but had not saved the file in two days?

jguidry-recurly: Would you believe that I actually have this fixed in an open buffer, but had not saved the file in two days? Thank you for the catch!

lib/recurly/webhook.rb
```diff
@@ -37,6 +37,7 @@ module Webhook
     autoload :SubscriptionResumedNotification,      'recurly/webhook/subscription_resumed_notification'
     autoload :CanceledSubscriptionNotification,     'recurly/webhook/canceled_subscription_notification'
     autoload :ScheduledSubscriptionPauseNotification,'recurly/webhook/scheduled_subscription_pause_notification'
+    autoload :ScheduledSubscriptionUpdateNotification, 'recurly/webhook/scheduled_subscription_update_notification'
```
joannasese: Super nit and feel free to ignore: move indentation over so all the `recurly/webhook/` paths line up.

jguidry-recurly: I went round about this one. It makes PR diff nastier, but since you know what's up, I'll push as another commit on here.

joannasese (approved):
👍

PR #727, opened 2021-09-24
Client library updates for Dunning Campaign API enhancements

amandamfielding:
* Add `dunning_campaign_id` to `Account`, `Invoice`, and `Plan` 
* PUT v2/dunning_campaigns/{dunning_campaign_id}/bulk_update
```
campaign = Recurly::DunningCampaign.find('id')
campaign.bulk_update(plan_codes)
```


sunnyw1212:
just a reminder that we'll need to wait for https://github.com/recurly/recurly-app/pull/17994 to be merged and implement

```
- Changed Account object response:
- Added dunning_campaign_id
- Changed POST v2/accounts and PUT v2/accounts/{account_code}
- Added <dunning_campaign_id> field
- Changed POST v2/purchases, POST /v2/purchases/preview and POST v2/purchases/authorize requests
- Added <dunning_campaign_id> field
- Changed POST v2/subscriptions and POST v2/subscriptions/preview requests
- Added <dunning_campaign_id> field
```

before we merge this cahnge

lib/recurly/dunning_campaign.rb
```diff
@@ -0,0 +1,33 @@
...
+      description
+      default
+      dunning_cycles
+      created_at
+      updated_at
+      deleted_at
+      plans
+    )
+
+    def bulk_update(plan_codes)
```
sunnyw1212, line 19: do we also need to support `plan_ids` here?

lib/recurly/dunning_campaign.rb
```diff
@@ -0,0 +1,33 @@
...
+      return false unless link? :bulk_update
```
sunnyw1212, line 20: can you explain what this is doing?

lib/recurly/dunning_campaign.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      builder = XML.new("<dunning_campaign/>")
+      node = builder.add_element("plan_codes")
+      plan_codes.each { |plan_code| node.add_element "plan_code", plan_code }
+
+      reload follow_link(:bulk_update, :body => builder.to_s)
```
sunnyw1212, line 26: can you explain what this is doing?

lib/recurly/resource.rb
```diff
@@ -608,7 +608,7 @@ def embedded!(root_index = false)
 
       def find_resource_class(name)
         resource_name = Helper.classify(name)
-        if Recurly.const_defined?(resource_name, false)
+        if Recurly.const_defined?(resource_name, false) && Recurly.const_get(resource_name, false).instance_of?(Class)
```
sunnyw1212, line 611: can you explain what this is doing?

judith: There is both an attribute `version` on the `DunningCycle` resource and a class `Version` this additional check makes sure that when it is looking for the resource class that it is also verifying that the constant is for the class and not an attribute with the same name.

lib/recurly/dunning_campaign.rb
judith: For API V2 the `plan_code` is the attribute for looking up a `Plan` resource. The `id` field is not exposed for plans in API V2.

lib/recurly/dunning_campaign.rb
judith: It is checking if there is an action link for `:bulk_update` as part of the response when instantiating the resource.

lib/recurly/dunning_campaign.rb
judith: It is making a request to the `href` with the `method` in the action link for that resource with the xml body for sending the plans and then creating a new instance of this resource from the xml response for that API request.

lib/recurly/invoice.rb
```diff
@@ -115,6 +115,7 @@ def invoice_number_with_prefix
       surcharge_in_cents
       tax_details
       billing_info_uuid
+      dunning_campaign_id
```
sunnyw1212, line 118: i think we only want dunning_campaign_id to be readable and NOT writable - do we need to indicate that differently somehow here?

judith: That should be handled in server side validation for the API request. No indication is needed in the client library.

judith (approved):

lib/recurly/dunning_cycle.rb
```diff
@@ -0,0 +1,18 @@
+module Recurly
+  class DunningCycle < Resource
+    define_attribute_methods %w(
+      type
+      applies_to_manual
```
judith: This should be `applies_to_manual_trial`.

judith (approved):

PR #781, opened 2022-08-02
Added support for ramp-pricing plans

csampson:
- Adds `pricing_model` attribute to `Recurly::Plan`, which can be either 'fixed' (for fixed-pricing plans) or 'ramp' (for ramp-pricing plans)
- Adds `ramp_intervals` attribute to `Recurly::Plan` (see final bullet-point)
- Adds `Recurly::PlanRampInterval` class, which can be used to initialize ramp_intervals for plans via the `ramp_intervals` attribute

### DIFF
--- a/spec/recurly/plan_spec.rb
+++ b/spec/recurly/plan_spec.rb
@@ -12,6 +12,7 @@ describe Plan do
       :setup_fee_in_cents        => 60_00,
       :plan_interval_length      => 1,
       :plan_interval_unit        => 'months',
+      :pricing_model             => 'fixed',
       :tax_exempt                => false,
       :revenue_schedule_type     => 'evenly',
       :avalara_transaction_type  => 600,
@@ -19,8 +20,9 @@ describe Plan do
     )
   }
 
-  it 'must serialize' do
-    plan.to_xml.must_equal <<XML.chomp
+  describe 'when pricing_model is fixed' do
+    it 'must serialize' do
+      plan.to_xml.must_equal <<XML.chomp
 <plan>\
 <accounting_code>gold_plan_acc_code</accounting_code>\
 <avalara_service_type>3</avalara_service_type>\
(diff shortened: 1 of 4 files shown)


spec/recurly/plan_spec.rb
```diff
@@ -41,6 +44,54 @@
 </unit_amount_in_cents>\
 </plan>
 XML
+    end
+  end
+
+  describe 'when pricing_model is ramp' do
```
Smolations, line 50: Is this really the only sort of testing done for the ruby lib? Just serialization from object->xml but not the other way around? And nothing dealing with the various requests?  🤔

Smolations (commented):
Looks pretty good. Just had a question around the tests.

danilocandido (approved):

danilocandido:
it looks great

spec/recurly/plan_spec.rb
csampson: Yeah, it's interesting. Some of the tests (especially those in `subscription_spec.rb`) do go into more detail, making stubbed requests and whatnot, but that feels a bit nonsensical to me.

I like the idea of the specs being as lean as the source file - the `Plan` class doesn't do much on its own, it merely wires things up for use by the `Resource` implementation (which itself is concerned with making http requests, etc.).

Smolations: Yea, I dunno, just doesn't personally give me a ton of confidence on the end-to-end bit of it. It just feels weird not to test every code path to ensure that bit of confidence. Like, it can serialize, but is it set up right to ingest xml? Does it work for initially ingested XML and updating based on a PUT update? Just...feels weird.  😅

csampson: > Like, it can serialize, but is it set up right to ingest xml?

That's basically the responsibility from `Resource#from_xml`, which has its own tests.

> Does it work for initially ingested XML and updating based on a PUT update?

That's the responsibility of `Resource#to_xml` (serializing the resource instance into an xml payload) and `Resource#save`, which also have their own tests.

PR #798, opened 2022-10-13
Update client library to support entitlements

lsfernandes92:
This PR aims to:
* Adds `CustomerPermission`, `Entitlement` and `ExternalSubscription` models.
* Adds `GET v2/accounts/{account_id}/entitlements` endpoint to this client library.

Sample code request:
```ruby
begin
  account = Recurly::Account.find('<ACCOUNT_CODE>')
  account.entitlements.find_each do |entitlement|
    puts "Entitlement: #{entitlement}"
  end
rescue Recurly::Error => e
  e.inspect
end
```


amandamfielding (approved):

lib/recurly/external_subscription.rb
```diff
@@ -0,0 +1,17 @@
+module Recurly
+  class ExternalSubscription < Resource
+
+    define_attribute_methods %w(
+      href
+    )
+
+    def href
+      @href
+    end
```
chrissrogers: We don't typically expose the href of a Resource as an attribute method, instead allowing the links to build associations.

lsfernandes92: Just for the record... I did this because the proper associate object wasn't being recovered and I was trying to maintain some default response pattern through the other client libraries. Although talking with Chris he explained that we're most interested in maintaining the pattern of the clients themselves. In order to pull up the associate objects from the `granted_by` node, the method `#reload` is given as an option to the merchants if they want to look for those objects. Like, for example:  `account.entitlements.first.granted_by.last.reload`.

PR #830, opened 2023-03-15
[external-invoices-endpoints] Add external charge and external invoice

Patrick-Duvall:
Add external_charge and external_invoice models and specs.


amandamfielding:
I think we need to add `has_many :external_invoices` to the external sub resource

amandamfielding:
I think we also need fixtures to test against

amandamfielding (approved):

PR #852, opened 2023-06-07
Multiple business entities

rlew421:
GET /business_entities
GET /business_entities/:id
GET /business_entities/:id/invoices

### DIFF
--- a/spec/recurly/account_spec.rb
+++ b/spec/recurly/account_spec.rb
@@ -433,6 +433,18 @@ XML
     end
   end
 
+  describe "override_business_entity_id" do
+    let(:account) {
+      stub_api_request :get, "accounts/abcdef1234567890", "accounts/show-200"
+      Account.find("abcdef1234567890")
+    }
+
+    it "should have an override business entity" do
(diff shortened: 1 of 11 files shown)


spec/recurly/account_spec.rb
```diff
@@ -433,6 +433,17 @@
...
 
+  describe "override_business_entity_id" do
+    let(:account) {
+      stub_api_request :get, "accounts/abcdef1234567890", "accounts/show-200"
+      Account.find("abcdef1234567890")
+    }
+
+    it "should have an override_business_entity_id" do
+      account.override_business_entity_id.must_equal("sbup2j0fx800")
+    end
```
Patrick-Duvall: For this and invoice_spec `account.business_entity.must_be_instance_of(BusinessEntity)` and `invoice.business_entity.must_be_instance_of(BusinessEntity)` might be more meaningful specs, since they show it has the associated object, and the `business_entity` behavior is specced in `business_entity_spec`, what do you think?

rlew421: That makes sense. I can make those changes

Patrick-Duvall (approved):

PR #873, opened 2024-01-22
Adding support for General Ledger Accounts

faiberrec:


lib/recurly/xml/rexml.rb
```diff
@@ -22,7 +22,7 @@ def each_element xpath = nil
       def each element = root
         element.each_element do |el|
           yield el
-          each el, &Proc.new
+          each el, &Proc.new { |e| yield e }
```
faiberrec, line 25: This was required to remove a bunch of noisy warnings.

PR #880, opened 2024-01-30
add get business entity

wsethbrown:


spec/fixtures/business_entities/index-200.xml
```diff
@@ -29,6 +29,8 @@ Link: <https://api.recurly.com/v2/business_entities?cursor=1234567890&per_page=2
 		</tax_address>
 		<default_vat_number nil="nil"></default_vat_number>
 		<default_registration_number nil="nil"></default_registration_number>
+    <default_liability_gl_account_id>12345</default_liability_gl_account_id>
+    <default_revenue_gl_account_id>56789</default_revenue_gl_account_id>
```
faiberrec, line 33: can we keep the same indentation plz?

wsethbrown: Ok, weird! I tried fixing this but the weird indents are not visible in my VSCode and I tried forcing new indents, but was unable to push anything as no change was detected.

Additionally, Smola said the indents looked right on his screen. I agree, would like to fix this but seems the issue is not universal for everyone?

faiberrec (approved):

faiberrec:
QA+1

PR #890, opened 2024-02-27
fix: Allow Resource to be compare to other object

malbert:
# Problem

The `Recurly::Resource` does not allow to compare to another object that does not respond to the `attributes` method.

### DIFF
--- a/lib/recurly/resource.rb
+++ b/lib/recurly/resource.rb
@@ -15,8 +15,8 @@ module Recurly
     end
 
     def ==(other_resource)
-      other.is_a?(Recurly::Resource) &&
-        attributes == other.attributes
+      other_resource.is_a?(Recurly::Resource) &&
+        attributes == other_resource.attributes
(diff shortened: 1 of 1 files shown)


douglasmiller:
Thank you for the contribution, @malbert. Could you fix the issues that this change introduces? You can run the tests locally using the test script: `./scripts/test`

lib/recurly/resource.rb
```diff
@@ -15,7 +15,8 @@ def requires_client?
     end
 
     def ==(other_resource)
-      self.attributes == other_resource.attributes
+      other.is_a?(Recurly::Resource) &&
```
douglasmiller: The method parameter is `other_resource`

malbert: Good catch!
Stupid error while renaming variables.
Thank you!

douglasmiller (approved):

PR #891, opened 2024-02-27
add revrec to purchases and adjustments

wsethbrown:
add revrec to purchases and adjustments


lib/recurly/invoice.rb
```diff
@@ -122,7 +122,7 @@ def invoice_number_with_prefix
       refundable_total_in_cents
       used_tax_service
       business_entity_id
-    )
+    ) + RevRec::PRODUCT_ATTRIBUTES
```
faiberrec: Why do we need the RevRec attributes in the invoice model?

I think this needs to be removed.

PR #898, opened 2024-04-25
Add subscription change proflex options

Patrick-Duvall:
Adds proration_settings to subscription attributes.

### DIFF
--- a/lib/recurly/subscription.rb
+++ b/lib/recurly/subscription.rb
@@ -109,6 +109,7 @@ module Recurly
       gateway_code
       transaction_type
       action_result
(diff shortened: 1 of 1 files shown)


lib/recurly/subscription.rb
```diff
@@ -109,6 +109,7 @@ class Subscription < Resource
       gateway_code
       transaction_type
       action_result
+      proration_settings
```
Patrick-Duvall, line 112: This behavior is already well tested in the Resource parent class specs.

andcoca (approved):

PR #943, opened 2025-11-27
Accept vertex_transaction_type fur purchase, adjustments and invoice

sgilrodriguez:


d14z-tech (dismissed):

sunnyw1212:
@sgilrodriguez not a hard requirement but thoughts about adding unit tests to ensure
- adjustments within a purchase request can have a vertex_transaction_type
- setting an invoice level vertex_transaction_type will override any adjustments with vertex_transaction_type set on them
?

spec/recurly/purchase_spec.rb
```diff
@@ -13,6 +13,103 @@
...
+          }
+        ]
+      )
+      xml = purchase.to_xml
+      # Purchase-level vertex_transaction_type
+      xml.must_match(/<purchase>.*<vertex_transaction_type>sale<\/vertex_transaction_type>.*<\/purchase>/m)
+      # Adjustment-level vertex_transaction_type
+      xml.must_match(/<adjustment>.*<vertex_transaction_type>lease<\/vertex_transaction_type>.*<\/adjustment>/m)
+    end
+  end
```
sunnyw1212, line 111: @sgilrodriguez do we actually return the top-level purchase vtt in the ruby lib response? we dont do it in dotnet because the purchase endpoint returns an invoice_collection

sunnyw1212: the postman api call to v2 purchases also returns just an invoice_collection

Screenshot 2025-12-03 at 1 26 34 PM

sgilrodriguez: The test only asserts the shape of the XML being produced, not the actual return values

sunnyw1212 (approved):

sunnyw1212:
devqa +1

